PR #16, opened 2022-02-11
ItemTypeName

LordMidas:
Added a method for adding itemtypes with names that are then shown below an item's name in the tooltip

### DIFF
--- a/scripts/!mods_preload/msu_mod_item.nut
+++ b/scripts/!mods_preload/msu_mod_item.nut
@@ -96,8 +96,7 @@ gt.MSU.modItem <- function ()
 		local getDescription = o.getDescription;
 		o.getDescription = function()
 		{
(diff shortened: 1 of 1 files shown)


scripts/!mods_preload/msu_mod_item.nut
```diff
@@ -2,7 +2,39 @@ local gt = this.getroottable();
...
+	
+	foreach (itemType in this.Const.Items.ItemType)
+	{
+		local name = "";
+		switch (itemType)
+		{
+			case this.Const.Items.ItemType.Legendary:
+				name = "Legendary Item";
+				break;
+
```
Enduriel, line 16: if we are using this type of `switch` syntax with an empty line, I know some of my code needs to be adjusted so it's all standardized

Enduriel (approved):
Big fan of this commit in general.

PR #17, opened 2022-02-11
A save-versioning serialization system for mods

LordMidas:


scripts/!mods_preload/msu_setup_save_version_system.nut
```diff
@@ -2,41 +2,44 @@ local gt = this.getroottable();
...
 gt.MSU.setupSaveVersionSystem <- function ()
 {
-	gt.MSU.Serialization.Mods <- [];
-	gt.MSU.Serialization.registerMod <- function( _mod, _ver )
-	{
-		if (typeof _mod != "string" || typeof _ver != "string")
+	gt.MSU.Serialization <- {
+		Mods = [],
+
+		function registerMod( _mod, _ver )
```
TaroEld: the Debug module already has a registerMod function, perhaps we need to consolidate them

scripts/!mods_preload/msu_serialization_system.nut
```diff
@@ -0,0 +1,73 @@
+local gt = this.getroottable();
+
+gt.MSU.setupSerializationSystem <- function ()
+{
+	gt.MSU.Serialization <- {
+		Mods = [],
+
+		function registerMod( _mod, _ver )
+		{
+			if (typeof _mod != "string" || typeof _ver != "string")
```
Enduriel, line 10: Any reason this isn't structured as ` a == 'string' && b == 'string'`? It would be much more logical that way imo than two inverses. with line 17 switched with the 3 in here obviously.

Enduriel:
As to the rest of this I'll take set up a standardized system for mod registration and we'll need to review all the systems that would interact with that (this, debug and settings for now)

scripts/!mods_preload/msu_serialization_system.nut
LordMidas: I want both the `_mod` i.e. mod name **and** the `_ver` to be strings. `&&` doesn't solve that. `&&` only returns false if both of them are not strings, but returns true when one is a string and the other isn't. What I want is `||` as it returns true only when both are strings.

Enduriel:
This is good to merge whenever you're happy with it from my side @LordMidas

PR #18, opened 2022-02-11
Skill container events

LordMidas:


scripts/!mods_preload/msu_mod_skill_container.nut
```diff
@@ -42,38 +42,7 @@ gt.MSU.modSkillContainer <- function ()
 	::mods_hookNewObject("skills/skill_container", function(o) {
 		o.m.BeforeSkillExecutedTile <- null;
 		o.m.IsExecutingMoveSkill <- false;
-
-		o.m.EventsDirectory <- {
```
TaroEld: Perhaps we could record this list somewhere because it's useful to see all the vanilla events at a glance

Enduriel (changes requested):
I agree with what @LordMidas said about reflecting on how much this actually benefits modders and users, I know I suggested this idea in the first place but now I'm uncertain that it would have a significant enough performance advantage to outweigh the increased complexity.

scripts/!mods_preload/msu_mod_skill_container.nut
```diff
@@ -2,9 +2,100 @@ local gt = this.getroottable();
...
+		],
+
+		function addEvent( _name, _func = null, _update = true, _aliveOnly = false )
+		{
+			this.EventsToAdd.push({
+				Name = _name,
+				Update = _update,
+				AliveOnly = _aliveOnly,
+				IsEmpty = _func == null,
+				Func = _func == null ? @() {} : _func,
```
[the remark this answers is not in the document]
LordMidas: Done

scripts/!mods_preload/msu_mod_skill_container.nut
LordMidas: Done

LordMidas:
We won't be implementing this for now. For now I will simply pull the addEvents function add that functionality. The EventsDictionary and such should be up for review later in case we want to do that.

PR #25, opened 2022-02-14
Systems system improvements

Enduriel:
Splits up Debug into two parts (the Log System and the static functions in Debug).
Changes Systems to require each mod to register with each system it wants to use.
Improves ModSettings handling of long mod names with a custom button.


scripts/!mods_preload/msu_aa.nut
```diff
@@ -8,28 +8,30 @@ gt.MSU.MSUModName <- MSUModName;
...
-	gt.MSU.setupUI();
-	delete gt.MSU.setupUI;
+
 	gt.MSU.setupSystems();
 	delete gt.MSU.setupSystems;
-
 	gt.MSU.setupModRegistry();
 	delete gt.MSU.setupModRegistry;
+
+	this.MSU.registerMod(MSUModName, "0.6.27", "Modding Standards and Utils");
```
LordMidas, line 23: I would like the name to be written as "Modding Standards & Utilities"

MSU-Documentation.md
```diff
@@ -848,15 +848,15 @@ Checks if debugging is enabled for mod `_modID` and flag `_flagID`.
...
+Convenience function for this.MSU.Systems.Log.isEnabledForMod.
 
 - `::printLog( _printText, _modID, _flagID = null)`, `::printWarning`, `::printError`
 
 Substitutes for `this.logInfo`, `this.logWarning` and `this.logError`. Prints the log as `_printText` if debugging is enabled for the mod id `_modID`. `_flagID` specifies a flag of the mod, and is set to default if left empty.
 
 `Other debugging tools`
 
-- `this.MSU.Debug.printStackTrace( _maxDepth = 0, _maxLen = 10, _advanced = false )`
+- `this.MSU.Systems.Log.printStackTrace( _maxDepth = 0, _maxLen = 10, _advanced = false )`
```
LordMidas, line 859: This isn't correct. The printStackTrace function is currently in `this.MSU.Debug` on your branch

scripts/!mods_preload/msu_debug.nut
```diff
@@ -206,43 +151,35 @@ gt.MSU.setupDebuggingUtils <- function()
...
-	::printError <- function( _arg,  _modID, _flagID = this.MSU.Debug.DefaultFlag)
+	::printError <- function( _arg,  _modID, _flagID = this.MSU.Systems.Log.DefaultFlag)
 	{
-		this.MSU.Debug.print(_arg, _modID, this.MSU.Debug.LogType.Error, _flagID);
+		this.MSU.Systems.Log.print(_arg, _modID, this.MSU.Systems.Log.LogType.Error, _flagID);
 	}
 
-	::isDebugEnabled <- function(_modID, _flagID = this.MSU.Debug.DefaultFlag){
-		return this.MSU.Debug.isEnabledForMod( _modID, _flagID)
+	::isDebugEnabled <- function(_modID, _flagID = this.MSU.Systems.Log.DefaultFlag){
```
LordMidas, line 173: If we are going with `this.MSU.Systems.Log` then the name of the shortcut needs to be updated perhaps to something like ``::isLogEnabled`

LordMidas (changes requested):
I will have to look deeper and then I'll give my thoughts on how we want to name the Debug/Logging system and where the static utility functions should be. It seems to me that the system should be called Debug, as it has room for expansion. If we call it Log system then it is very restricted.

TaroEld (approved):

PR #31, opened 2022-03-08
Make enum_settings have the same param order as other settings

Enduriel:
Technically a breaking change, noticed this while working on eimo and it bothered me a lot so I wanna fix it before release. Let me know if you have any issues with it.


TaroEld (approved):
lgtm

PR #32, opened 2022-03-09
Add mirror of mod-settings in JS code using MSUConnection

Enduriel:
In order to have a setting update the JS code when it has a new value set, settings elements are now aware of the panelid of the panel they are in. I tried to think of ways around this but the only one I could think of was making settings not be objects and instead do all the computation within panels which imo isn't really an option anymore.

I'm very open to other suggestions though as I'm not really that happy with this solution.

### DIFF
--- a/msu/systems/mod_settings/settings_page.nut
+++ b/msu/systems/mod_settings/settings_page.nut
@@ -3,6 +3,7 @@ this.MSU.Class.SettingsPage <- class
 	Name = null;
 	ID = null;
 	Settings = null;
+	PanelID = null;
 
 	constructor( _id, _name = null )
 	{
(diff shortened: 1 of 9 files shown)


TaroEld:
I think it's perfectly reasonable to have settings know what panel they're in. In fact I'd go even further and just set a parent variable equal to the parent instance. Ie page has panel as parent, setting has page as parent, then getPanelID() functions to grab the ID from the respective parent.

Enduriel:
Yeah I was considering that too, and that is probably a better long term solution.

msu/systems/mod_settings/settings_page.nut
```diff
@@ -18,9 +19,20 @@ this.MSU.Class.SettingsPage <- class
...
+	function setParent( _parent )
+	{
+		this.Parent = _parent;
+	}
+
+	function getParent()
+	{
+		return this.Parent;
+	}
+
```
TaroEld, line 35: Probably also add getPanelID() function to this

PR #52, opened 2022-03-17
JS Keybinds Settings Improvements

Enduriel:


ui/mods/msu/css/misc.css
```diff
@@ -1,10 +1,29 @@
...
+{
+	outline-color: green;
+}
+
+.blue
+{
+	outline-color: blue;
+}
+
+
```
TaroEld, line 26: those should have more distinct class names such as outline-black

PR #56, opened 2022-03-18
feat: add way for modders to modify blueprints under specified conditions

LordMidas:
This method allows modders to change blueprints depending on conditions which aren't available when the game is starting up e.g. the current scenario being played.


Enduriel:
Like I said, I'm not a fan of this interface but I wanna get a release now so I think we shouldn't worry about this for 1.0

PR #59, opened 2022-03-18
Move from this to :: for all relevant references to roottable

Enduriel:
I know @LordMidas is still uncertain about this and I'm happy to discuss it more but we're moving too slowly right now and we just need to get stuff done.


TaroEld (commented):
I am fine with these changes, it is indeed more explicit

PR #60, opened 2022-03-18
docs: remove documentation, clear changelog and add readme

Enduriel:
Readme nexusmods link will need to be added in the future
Readme should also get a proof read by people


README.md
```diff
@@ -0,0 +1,13 @@
+# Modding Standards & Utilities
+is a modding library for the game [Battle Brothers](https://store.steampowered.com/app/365360/Battle_Brothers/)
+allowing for better mod compatibility
+and providing commong functions for mods.
```
LordMidas, line 4: common*

README.md
```diff
@@ -0,0 +1,13 @@
...
+
```
LordMidas, line 5: I would add some more flashy info e.g. "provides groundbreaking modding features that significantly enhance mod compatibility and allows modders to write faster, cleaner, and more sustainable code" etc.

LordMidas (changes requested):

README.md
Enduriel: I wouldn't wanna lay it on _that_ thick, but I see where you're coming from.

Enduriel: Actually I take it back, I think it's fine. I think we can flatter ourselves in the nexusmods description where the users can see it, this is meant to be fairly minimalistic imo (at least for now).

LordMidas: Yeah, as long as MSU comes across as something important. I'd still add some adjective like "significantly better" rather than just "better".

PR #62, opened 2022-03-19
feat: add LevelUpsSpent variable in player

LordMidas:


Enduriel:
What is the point of this?

LordMidas:
> What is the point of this?

In PTR I have a perk that does something depending on how many prior LevelUps the character has spent. Conceivably, this info could be useful for other perks too.

Enduriel:
How is this different from just checking the level of the player?

LordMidas:
> How is this different from just checking the level of the player?

Because when you level-up you haven't yet spent the level-up from that level. This is exacerbated when a character levels up multiple levels in a single go. This is crucial for a perk in PTR that alters the character's talents and level-ups for future level-ups. That perk needs to know the exact number of level-ups spent.

msu/hooks/entity/tactical/player.nut
```diff
@@ -30,4 +39,18 @@
...
+	{
+		this.getFlags().set("LevelUpsSpent", this.m.LevelUpsSpent);
+		onSerialize(_out);
+	}
+
+	local onDeserialize = o.onDeserialize;
+	o.onDeserialize = function( _in )
+	{
+		onDeserialize(_in);
+		this.m.LevelUpsSpent = this.getFlags().get("LevelUpsSpent");
```
Enduriel, line 54: I just realized this will default to false which is probably not ideal, you should add a `has` check and set it to 0 if false.

msu/hooks/entity/tactical/player.nut
```diff
@@ -30,4 +39,18 @@
 			}
 		}
 	}
+
+	local onSerialize = o.onSerialize;
+	o.onSerialize = function( _out )
+	{
+		this.getFlags().set("LevelUpsSpent", this.m.LevelUpsSpent);
+		onSerialize(_out);
```
Enduriel, line 47: Also imo it's good practice to remove flags after you serialize with them just to keep everything clean

msu/hooks/entity/tactical/player.nut
```diff
@@ -1,4 +1,13 @@
 ::mods_hookExactClass("entity/tactical/player", function(o) {
+	o.m.LevelUpsSpent <- 0;
+
+	local setAttributeLevelUpValues = o.setAttributeLevelUpValues;
+	o.setAttributeLevelUpValues = function( _v )
```
Enduriel, line 5: Is this the vanilla parameter name again?

Enduriel (changes requested):

msu/hooks/entity/tactical/player.nut
LordMidas: Yes

msu/hooks/entity/tactical/player.nut
LordMidas: Good catch. Will do.

msu/hooks/entity/tactical/player.nut
LordMidas: Just to make sure that I understand correctly, you mean that the flag used for serialization should be removed, so that the `tag_collection` doesn't have that entry anymore while the play session is still running? If yes, then I guess that's definitely cleaner. Although, I don't see how leaving the flag in could cause any trouble. Any examples?

Enduriel: In this case it shouldn't cause issues, but if you are for example using the existence of a flag to share information, not removing it after serialization will cause issues. Therefore imo it's best to just follow this convention everywhere to make sure you don't run into that issue, as well as saving on a bit of memory :P

PR #63, opened 2022-03-19
On update level tweaks

LordMidas:


Enduriel:
Why is this necessary? I feel this this will definitely break stuff

LordMidas:
> Why is this necessary? I feel this this will definitely break stuff

Yes this will break stuff which is why I'm including it in 1.0 along with several other changes that are breaking things. The primary reason I want to do this is because in `onUpdateLevel` the modder almost always wants to check `this.getContainer().getActoer().getLevel()` which can be avoided if the function just passes that as a parameter i.e. it makes life easier.

Enduriel:
No, I don't like this. I don't think we should be breaking compatibility for something that's readily available from the scope of the function.

Passing arguments for stuff that is already available to the function is bad practice anyway imo.

PR #64, opened 2022-03-19
feat: add onEnterSettlement event for skills

LordMidas:


msu/hooks/entity/world/settlement.nut
```diff
@@ -0,0 +1,18 @@
...
+			local roster = this.World.getPlayerRoster().getAll();
+			foreach (bro in roster)
+			{
+				bro.getSkills().onEnterSettlement(this);
+			}
+		}
+
+		return ret;
+	}
+});
```
Enduriel, line 18: someone doesn't have editorconfig set up

Enduriel (approved):

Enduriel (changes requested):
That was actually supposed to be an approve if you add that missing newline

msu/hooks/entity/world/settlement.nut
LordMidas: Hmm... I wonder who that must be :-s

PR #67, opened 2022-03-20
Weighted container tweaks

LordMidas:


msu/classes/weighted_container.nut
```diff
@@ -35,6 +36,34 @@
...
+	}
+
+	function getProbability( _item )
+	{
+		local idx = this.find(_item);
+		if (idx == null) throw ::MSU.Exception.KeyNotFound;
+		return this.Array[idx][0].tofloat() / this.Total;
+	}
+
+	function sort( _function )
```
LordMidas, line 61: An alternative implementation of this function could be. However, in this case I am not sure if the default sorting i.e. when _function is null should be by weight (index 0) or by value (index 1).
```
function sort( _function = null )
{
	if (_function != null)
	{
		::MSU.requireFunction(_function);
		this.Array.sort(_function);
	}
	else
	{
		this.Array.sort(@(a, b) a[0] <=> b[0]);
	}
}
```

msu/classes/weighted_container.nut
```diff
@@ -35,6 +36,34 @@
...
+		{
+			if (item[1] == _item) return i;
+		}
+
+		return null;
+	}
+
+	function remove( _idx )
+	{
+		this.Array.remove(_idx);
```
LordMidas, line 51: Should probably be `return this.Array.remove(_idx)` or `return this.Array.remove(_idx)[1]`. Which one is better?

msu/classes/weighted_container.nut
```diff
@@ -6,6 +6,7 @@
 	constructor( _array = null )
 	{
 		if (_array == null) _array = [];
+		::MSU.requireArray(_array);
```
Enduriel, line 9: I specifically removed this because it's unnecessary.

msu/classes/weighted_container.nut
Enduriel: I don't think this needs to implement sort, I think remove is reasonable because I can see a use-case for it, but this shouldn't even expose that it's a sorted array. If you're trying to sort this you're trying to use it for something it wasn't designed for imo

PR #69, opened 2022-03-20
Weighted container tweaks 2

LordMidas:


Enduriel:
This PR isn't set up correctly so reviewing it is a massive pain, would be create if you could recreate it with the right branches

LordMidas:
Closing because of wrong target branch.

PR #70, opened 2022-03-21
refactor: remove hook on updateLevel of player and move the call to onUpdateLevel to a starting_scenario hook

LordMidas:


msu/hooks/scenarios/world/starting_scenario.nut
```diff
@@ -9,3 +9,22 @@
...
 });
+
+::mods_hookDescendants("scenarios/world/starting_scenario", function(o) {
+	if ("onUpdateLevel" in o)
+	{
+		local onUpdateLevel = o.onUpdateLevel;
+		o.onUpdateLevel = function( _bro )
+		{
+			onUpdateLevel(_bro);
+			_bro.getSkills().onUpdateLevel();
```
LordMidas, line 20: Reverse the order to be consistent with how it'll behave with scenarios which don't have onUpdateLevel defined in them as MSU hook runs before everything else.

msu/hooks/scenarios/world/starting_scenario.nut
```diff
@@ -9,3 +9,22 @@
...
+		}
+	}
+	else
+	{
+		o.onUpdateLevel <- function( _bro )
+		{
+			_bro.getSkills().onUpdateLevel();
+		}
+	}
```
Enduriel, line 29: This still isn't able to properly handle situations were 2 scenarios inherit from each other, and the one that inherits doesn't have an onUpdateLevel function.

Enduriel (changes requested):
We need to account for scenarios inheriting from each other, right now this breaks in those situations

msu/hooks/scenarios/world/starting_scenario.nut
LordMidas: I don't understand. If you are playing a scenario that inherits, it will call its parent's onUpdateLevel function which, if it did exist, will have the call on the skills onUpdateLevel. It will work.

LordMidas: Check now

PR #72, opened 2022-03-21
feat: remove the hook on getDefaultTooltip and add getRangedTooltip

LordMidas:
This is better, imo, than a `getDefaultRangedTooltip` function for multiple reasons:
1. Shorter name.
2. Flexibility. This function does not assume that you need the `DefaultTooltip` as well.

So the way this function would be used in a skill by a modder is the following:
```
function getTooltip()
{
	local tooltip = this.getDefaultTooltip();
	tooltip.extend(this.getRangedTooltip());
	return tooltip;
}
```
or a one-liner:
```
function getTooltip()
{
	return this.getDefaultTooltip().extend(this.getRangedTooltip());
}
```


TaroEld (commented):
lgtm

Enduriel (approved):
Seems reasonable to me

PR #73, opened 2022-03-22
Close Issue #14

LordMidas:
refactor!: remove getTotalArmorStaminaModifier from actor and add a substitute function in item_container

The new function accepts either an array of slots or a single slot. By default it returns the total StaminaModifier of items in all slots. It has built-in handling for items which may not have the getStaminaModifier() function in them.

The rationale behind removing the function from actor and adding to item_container is that StaminaModifier is a purely item property and while perks such as Brawny affect how much stamina the actor loses due to his items, the StaminaModifier of the items remains static. Therefore, keeping this function in the actor is misleading.

### DIFF
--- a/msu/hooks/items/item_container.nut
+++ b/msu/hooks/items/item_container.nut
@@ -39,6 +39,43 @@
 		this.m.Actor.getSkills().update();
 	}
 
+	o.getStaminaModifier <- function( _slots = null )
+	{
+		if (_slots == null)
+		{
+			 _slots = clone ::Const.ItemSlot;
+			 delete _slots.None;
+			 delete _slots.COUNT;
+		}
+		else
+		{
+			if (typeof _slots == "integer") _slots = [_slots];
+		}
+
+		local ret = 0;
+
+		foreach (slot in _slots)
+		{
+			local items = this.getAllItemsAtSlot(slot);
+			foreach (item in items)
+			{
+				// Avoid exceptions for items which don't have getStaminaModifier()
+				// This handles cases such as Quivers in ammo slot which don't have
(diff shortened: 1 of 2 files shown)


msu/hooks/items/item_container.nut
```diff
@@ -39,6 +39,43 @@
...
+				// Avoid exceptions for items which don't have getStaminaModifier()
+				// This handles cases such as Quivers in ammo slot which don't have
+				// this function in vanilla but do in mods like Legends
+				try
+				{
+					ret += item.getStaminaModifier();
+				}
+				catch (error)
+				{
+					::MSU.Mod.Debug.printError(error + " -- for item: " + item.getID(), "debug");
```
LordMidas, line 71: I have half a mind to leave the catch(error) block empty and print nothing to the log, because I think these messages will mostly be stuff that we can't do anything about and will just clog up the debug log instead of giving any useful info. They don't represent any problem in MSU so I think this printError should be removed.

TaroEld, line 71: Why try catch instead of checking `if ("getStaminaModifier" in item)`

Enduriel: Agree with Taro but overall a fan of this commit

Enduriel (changes requested):

msu/hooks/items/item_container.nut
LordMidas: Because of WeakTableRef shit

LordMidas: Will redo this based on MSU.isIn

Enduriel (approved):
lgtm

PR #74, opened 2022-03-22
feat: add MSU.isKindOf function which handles both BB tables and WeakTableRef instances

LordMidas:
The vanilla BB isKindOf function only works with BB tables. If you pass a WeakTableRef instance to it, it fails. The MSU version handles both.

### DIFF
--- a/msu/utilities/misc.nut
+++ b/msu/utilities/misc.nut
@@ -37,3 +37,23 @@
 			throw ::MSU.Exception.InvalidType(_object);
 	}
 }
+
+::MSU.isKindOf <- function( _object, _className )
+{
+	local obj = _object;
+	if (typeof obj == "instance")
+	{
+		if (obj instanceof ::WeakTableRef)
+		{
+			if (obj.isNull())
+			{
+				::printError("The table inside the WeakTableRef instance is null");
+				throw ::MSU.Exception.KeyNotFound(_className);
+			}
+			obj = obj.get();
+		}
+		else throw ::MSU.Exception.InvalidType(_object);
+	}
+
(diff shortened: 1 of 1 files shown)


TaroEld (commented):
Should it also handle other classes? Ie do a _object instanceOf _className comparison.

Enduriel:
I'm pretty sure this breaks the original reference, you have to create a new variable to make this work.

Enduriel (commented):
Need to fix it overwriting the value of _object.

Enduriel (changes requested):
Meant to be a request changes

LordMidas:
> Should it also handle other classes? Ie do a _object instanceOf _className comparison.

I don't think that's necessary because in this case _className will not be a string. And when you know that you are passing a non-string _className then you can just use the instanceof keyword instead of calling MSU.isKindOf.

Enduriel:
Agree with Midas there, those as separate things imo, and instanceof is actually quite convenient imo

msu/utilities/misc.nut
```diff
@@ -37,3 +37,23 @@
...
+	local obj = _object;
+	if (typeof obj == "instance")
+	{
+		if (obj instanceof ::WeakTableRef)
+		{
+			if (obj.isNull())
+			{
+				::printError("The table inside the WeakTableRef instance is null");
+				throw ::MSU.Exception.KeyNotFound(_className);
+			}
```
Enduriel, line 52: Actually is this even necessary? wouldn't it throw the right error inside of isKindOf even if we didn't do this?

PR #77, opened 2022-03-22
Key getting utils

LordMidas:
The idea of this PR is to restrict the behavior of the `isIn` function as much as possible to how the default `in` keyword works, while implementing separate functions for BB specific classes to look up their chain of "inheritance".

### DIFF
--- a/msu/utilities/misc.nut
+++ b/msu/utilities/misc.nut
@@ -29,12 +29,10 @@
 
 	_object = _object.m;
 
-	if (_key in _object) return _object[_key];
-
 	do
 	{
-		_object = _object.getdelegate();
 		if (_key in _object) return _object[_key];
+		_object = _object.getdelegate();
 	}
 	while(_object != null);
 
@@ -57,12 +55,10 @@
 		else throw ::MSU.Exception.InvalidType(_object);
 	}
 
-	if (_key in _object) return _object[_key];
-
 	do
 	{
-		_object = _object.getdelegate();
 		if (_key in _object) return _object[_key];
+		_object = _object.getdelegate();
 	}
 	while(_object != null);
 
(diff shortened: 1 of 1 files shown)


msu/utilities/misc.nut
```diff
@@ -1,59 +1,99 @@
 ::MSU.isNull <- function( _object )
 {
-	if (_object == null)
+	return _object == null || (typeof _object == "instance" && _object instanceof ::WeakTableRef && _object.isNull());
+}
+
+::MSU.getField <- function( _key, _object )
```
Enduriel, line 6: Why are these params in this order?

Enduriel (changes requested):
Other than that lgtm

msu/utilities/misc.nut
TaroEld: 'get this key from that object' makes sense to me. However, it's the other way around for the hook functions, we should probably use that then.

LordMidas: This order makes much more sense so I'm really annoyed if we have to end up changing it just because of modhooks. We should encourage the people to use these functions anyway over the modhooks ones (although the modhooks ones are more performant due to avoiding a typeof check in cases where you are sure the object is a table). But yeah I can understand that it will create confusion and likely errors when modders use both MSU and modhooks functions. What do you say we do here then? Use the shit modhooks order?

Enduriel: I think it's more intuitive the other way tbh, because to me this replaces `_object._key` but I don't care that much. If you're not convinced just leave it as is and lets go with it.

LordMidas: Ok we'll go with `(_object, _key)`

LordMidas: Actually on second thought, the idea of these functions is to be used with `::MSU.isIn ( _key, _object )` and there the order is in that order. So it will be a bit weird if someone does:
```squirrel
if (::MSU.isIn(myKey, myObject)
{
   local member = ::MSU.getMember(myObject, myKey) // rather than (myKey, myObject)
}
```
I really can't decide which is better, so @TaroEld and @Enduriel please vote.

PR #78, opened 2022-03-23
MSU.Error and refactor code accordingly

LordMidas:
This accomplishes two things:

1. Keeps true Exceptions separate from info-containing Error messages.
2. Avoids copy pasting strings, as Error messages are contained in MSU.Error and we only need to edit them at one place if we want to change something.

**PR implementation:**
In this PR, it requires two lines of code wherever you wanna throw an error and exception:
```squirrel
::logError(MSU.Error.ModNotRegistered(_key));
throw ::MSU.Exception.KeyNotFound(_key);
```
**Alternative:**
An alternative method could be to implement a `::MSU.exceptionWithError( _exception, _error = null )` function and use it as follows. The name of this function could be shortened to something like `::MSU.exception` or `::MSU.exEr` etc.
```squirrel
throw ::MSU.exceptionWithError(::MSU.Exception.KeyNotFound(_key), ::MSU.Error.ModNotRegistered(_key));
```
Let me know what you guys prefer and, if you prefer the alternative, what name you would like for the function.

### DIFF
--- a/msu/systems/debug/debug_system.nut
+++ b/msu/systems/debug/debug_system.nut
@@ -48,7 +48,8 @@
 	{
 		if (!(_modID in this.Mods))
 		{
-			throw ::MSU.Exception.ModNotRegistered(_modID);
+			::logError(::MSU.Error.ModNotRegistered(_modID));
+			throw ::MSU.Exception.KeyNotFound(_modID);
 		}
 		this.Mods[_modID][_flagID] <- _flagBool;
 		if (_flagBool == true)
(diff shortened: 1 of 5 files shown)


TaroEld (approved):
I do prefer the way it is in this PR, keeps things nice and separate.

Enduriel:
I like this, but the terminology seems to be misleading. I'd prefer something like `::MSU.ExceptionInfo` for the additional information.

Enduriel (changes requested):
I like this, but the terminology seems to be misleading. I'd prefer something like `::MSU.ExceptionInfo` for the additional information.

LordMidas:
> I like this, but the terminology seems to be misleading. I'd prefer something like `::MSU.ExceptionInfo` for the additional information.

If we are going with the 2-line method then `::MSU.Error` is better imo as it is printed via `::logError` and its only job is to serve as a repository of frequently thrown errors.

Enduriel:
Ya okay, it's not a big deal anyway.

PR #79, opened 2022-03-23
Utility functions for Array/Table serialization

LordMidas:
See the code block below for explanation of the use case. If you guys prefer, **we can limit the serialization to only value types** i.e. integer, boolean, string for now until we are confident we wanna add the table/array stuff (because that can lead to issues if it is an array containing BB classes for example). If we implement this, we can simplify the (de)serialize functions of WeightedContainer to use these utility functions. In cases where the **array/table contains mixed value types**, these functions make life even easier. For example **see PR #81** where I have implemented a serialization system for OrderedMap using these utility functions.
```squirrel
function onSerialize( _out )
{
	_out.writeU32(this.Array.len());
	foreach (item in this.Array)
	{
		_out.writeU32(item); // if it is integer
		 // otherwise use the appropraite writeBool, writeString etc.
	}
}

function onDeserialize( _in )
{
	this.Array = array(_in.readU32(), null);
	for (local i = 0; i < size; i++)
	{
		this.Array[i] = _in.readU32(); // or the appropriate readBool, readString etc.
	}
}

// Instead of the above, with these utility functions you would do:

function onSerialize( _out )
{
	::MSU.serializeObject(this.Array, _out);
}

function onDeserialize( _in )
{
	this.Array = ::MSU.deserializeObject(_in);
}
```

### DIFF
--- a/msu/systems/serialization/serialization_system.nut
+++ b/msu/systems/serialization/serialization_system.nut
@@ -1,6 +1,14 @@
 ::MSU.Class.SerializationSystem <- class extends ::MSU.Class.System
 {
 	Mods = null;
+	static DataType = {
+		Integer = 0,
+		Float = 1,
+		Boolean = 2,
+		String = 3,
+		Array = 4,
+		Table = 5
+	};
 
 	constructor()
 	{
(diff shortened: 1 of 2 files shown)


TaroEld (commented):
Looks good to me. Needs testing to make sure that there's no subtle bug, though.

msu/utilities/misc.nut
```diff
@@ -57,3 +57,102 @@
...
+				{
+					_out.writeString("I");
+					_out.writeI32(element);
+				}
+				else
+				{
+					_out.writeString("U");
+					_out.writeU32(element);
+				}
+				break;
```
Enduriel, line 90: I think it'd be cool if we could use a boolean here since this writes to raw bytes and that actually makes a difference in the size of the final file.

Enduriel: Or at least a very short int if need be

msu/utilities/misc.nut
```diff
@@ -57,3 +57,102 @@
...
+				_out.writeBool(element);
+				break;
+
+			case "array":
+			case "table":
+				::MSU.serializeObject(element, _out);
+				break;
+
+			default:
+				throw ::MSU.Exception.InvalidType(element);
```
Enduriel, line 109: I'd wanna look into dealing with BB classes cuz in principle I don't see why we shouldn't be able to handle them, but I guess that can be done for 1.1.0

Enduriel (changes requested):

msu/utilities/misc.nut
LordMidas: Will go with bool

msu/utilities/misc.nut
LordMidas: So should we limit these functions to serialization of value types only for 1.0.0?

Enduriel: I think it's fine to say we serialize standard tables, arrays and value types for now and then we can see where we go from there. I think this system works fine.

Enduriel:
Maybe having a utility function to convert types to ints and vice versa would be nice? Seems like somethinf quite error prone otherwise, or a table instead of a function but just something that does it it one place. Rather than having to remember what is what

TaroEld (commented):
Instead of using magic numbers for the types, instead codify them; for example _out.writeU8(this.MSU.Serialization.TypeInt.Integer) or _out.writeU8(this.MSU.Serialization.TypeInt.Table)

TaroEld (approved):
LGTM

msu/systems/serialization/serialization_system.nut
```diff
@@ -14,4 +22,116 @@
...
+		local type = _in.readU8();
+		local isTable = type == this.DataType.Table;
+		local size = _in.readU32();
+
+		local ret = isTable ? {} : array(size, null);
+
+		for (local i = 0; i < size; i++)
+		{
+			local key = isTable ? _in.readString() : null;
+			local dataType = _in.readU8();
```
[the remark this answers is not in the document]
LordMidas: I've saved it because of the throw on line 128 where I need the data type.

PR #80, opened 2022-03-23
Weighted container refactor

LordMidas:
As explained in PR #81, shouldn't the `map` function look like this?
```squirrel
function map( _function )
{
	// _function (_item, _weight)

	local ret = ::MSU.Class.WeightedContainer();
	foreach (pair in this.Array)
	{
		ret.add(_function(pair[1], pair[0]));
	}
	return ret;
}
```
instead of the current:
```squirrel
function map( _function )
{
	// _function (_item, _weight)

	return ::MSU.Class.WeightedContainer(clone this.Array).apply(_function);
}
```

### DIFF
--- a/msu/classes/weighted_container.nut
+++ b/msu/classes/weighted_container.nut
@@ -157,6 +157,23 @@
 		this.Array.clear();
 	}
 
+	function top()
+	{
+		local weight = 0;
+		local ret;
+
+		foreach (pair in this.Array)
+		{
+			if (pair[0] > weight)
+			{
+				weight = pair[0];
+				ret = pair[1];
+			}
+		}
+
+		return ret;
+	}
+
 	function roll()
 	{
 		local roll = ::Math.rand(1, this.Total);
--- a/msu/classes/weighted_container.nut
+++ b/msu/classes/weighted_container.nut
@@ -118,10 +118,12 @@
 
 	function apply( _function )
 	{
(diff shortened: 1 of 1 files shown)


msu/classes/weighted_container.nut
```diff
@@ -118,32 +119,61 @@
 
 	function apply( _function )
 	{
+		// _function (_item, _weight)
+
 		foreach (i, pair in this.Array)
 		{
 			this.ApplyIdx = i;
-			_function(pair[1]);
+			_function(pair[1], pair[0]);
```
Enduriel, line 127: yeah I'd just go with passing the two, if they need a reference to the array they can just pass it themselves when creating the closure

msu/classes/weighted_container.nut
```diff
@@ -118,32 +119,61 @@
...
 
 		this.ApplyIdx = null;
 	}
 
 	function map( _function )
 	{
-		return ::MSU.Class.OrderedMap(clone this.Array).apply(_function);
+		// _function (_item, _weight)
+
+		return ::MSU.Class.WeightedContainer(clone this.Array).apply(_function);
```
Enduriel, line 137: Yeah change this like we discuss in #81.

Enduriel (changes requested):

Enduriel (approved):
lgtm

PR #81, opened 2022-03-23
Ordered map tweaks

LordMidas:
My main question is about the `map` function. Looking at the description of the `map` function in squirrel documentation it says:
> This method applies the supplied function to every item in the target array, one by one. The results are placed in a new array created by map() and returned by it. The new array will be the same size as the target array.
Contrast map() with [apply()](https://developer.electricimp.com/squirrel/array/apply) which changes the array’s items in situ.
The passed function must include a single parameter into which each item in the target array is passed. It is the function’s job to make sure the array values passed are appropriate to the changes it will make, if any.

and

> Creates a new array of the same size. For each element in the original array invokes the function ‘func’ and assigns the return value of the function to the corresponding element of the newly created array. Provided func can accept up to 3 arguments: array item value (required), array item index (optional), reference to array itself (optional).

So basically it applies the _function to the elements of the current array and adds the RESULT i.e. return value of that function to the new array. We are not doing this, we are instead adding the elements and then applying the function on them. This is not the same.

So what should we do? (PS: The same question applies to the map function of the WeightedContainer).

**Second question**
How can I allow the passed _function to have 1 required and up to 2 additional optional arguments? How do I handle that e.g. in the `apply` function?

### DIFF
--- a/msu/classes/ordered_map.nut
+++ b/msu/classes/ordered_map.nut
@@ -117,8 +117,11 @@
 	{
 		// _function (_key, _val, _idx)
 
-		local ret = this.filter(@(key, val) return true);
-		ret.apply(_function);
+		local ret = ::MSU.Class.OrderedMap();
+		foreach (i, key in this.Array)
+		{
+			ret[key] <- _function(key, this.Table[key], i);
+		}
 		return ret;
 	}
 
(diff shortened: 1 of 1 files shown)


Enduriel:
As to 1, I think you're right and we should duplicate the code to do those slightly different things even though it's very similar.
As to 2, from I gather that's not possible without some sort of try-catch trickery

PR #89, opened 2022-04-08
Change skill DamageType to be a WeightedContainer

LordMidas:


Enduriel (changes requested):

Enduriel:
Other than that lgtm and I really like how clean it is ngl

Enduriel (approved):
Yeah okay, leaving it like this works too idrm

PR #92, opened 2022-04-10
Auto register

TaroEld:
Automatically registers mods when they're created/register with MSU.
I also added two little helper functions to the debug addon to quickly enable/disable debug with the default flag since the constructor parameter is gone.

### DIFF
--- a/msu/msu_mod/setup_msu_mod.nut
+++ b/msu/msu_mod/setup_msu_mod.nut
@@ -1,0 +1,1 @@
+::MSU.Mod <- ::MSU.registerMod(::MSU.ID, ::MSU.Version, ::MSU.Name);
--- a/msu/msu_mod/setup_msu_mod.nut
+++ b/msu/msu_mod/setup_msu_mod.nut
@@ -1,1 +1,1 @@
-::MSU.Mod <- ::MSU.registerMod(::MSU.ID, ::MSU.Version, ::MSU.Name);
+::MSU.Mod <- ::MSU.Class.Mod(::MSU.ID, ::MSU.Version, ::MSU.Name);
--- a/msu/systems/debug/debug_mod_addon.nut
+++ b/msu/systems/debug/debug_mod_addon.nut
@@ -20,6 +20,16 @@
 		return ::MSU.System.Debug.isFullDebugForMod(this.Mod.getID());
 	}
 
+	function enable()
+	{
+		::MSU.System.Debug.setFlag(this.Mod.getID(), "default", true);
+	}
+
+	function disable()
+	{
+		::MSU.System.Debug.setFlag(this.Mod.getID(), "default", false);
+	}
+
 	function isEnabled( _flagID = "default" )
 	{
 		return ::MSU.System.Debug.isEnabledForMod(this.Mod.getID(), _flagID);
--- a/msu/systems/debug/debug_mod_addon.nut
+++ b/msu/systems/debug/debug_mod_addon.nut
@@ -1,8 +1,13 @@
 ::MSU.Class.DebugModAddon <- class extends ::MSU.Class.SystemModAddon
 {
-	function setFlags( _flagTable )
+	function enable()
 	{
-		::MSU.System.Debug.setFlags(this.Mod.getID(), _flagTable);
+		::MSU.System.Debug.setAllFlags(this.Mod.getID(), true);
+	}
+
+	function disable()
+	{
+		::MSU.System.Debug.setAllFlags(this.Mod.getID(), false);
 	}
 
 	function setFlag( _flagID, _flagBool )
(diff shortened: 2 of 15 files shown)


msu/systems/debug/debug_mod_addon.nut
```diff
@@ -20,6 +20,16 @@
 		return ::MSU.System.Debug.isFullDebugForMod(this.Mod.getID());
 	}
 
+	function enable()
```
LordMidas, line 23: As this only sets the Default flag, I'd suggest renaming the function to something like `setDefaultFlag`. Because the name `enable` seems to imply to me what I suggested some time ago i.e. enable/disable the debug system for a mod keeping the flags in their current state.

And I still believe that a Boolean to enable/disable the Debug system while keeping flag values intact is a useful feature, as it allows e.g. a modder to make a setting that can be toggled on/off by a user to enable/disable debug. I don't see how All or None is better than that. A good example can be whereby a Modder creates a "Debug" page in his settings panel where all the different flag types can be toggled on/off. And when a user is facing a certain problem, the modder can simply ask the user to toggle a certain flag on/off while enabling/disabling the Debug system as necessary to prevent log spam.

msu/systems/mod.nut
```diff
@@ -67,17 +64,9 @@
...
-		if (vargv == null) vargv = [];
-		local args = [_system, this];
-		args.extend(vargv);
-		_system.registerMod.acall(args);
-	}
-
 	function tostring()
 	{
-		return format("Mod %s, Versions %s\n Registered Systems:\nDebug: %s\nKeybinds: %s\nModSettings: %s", this.getID(), this.getVersionString(), (this.Debug == ::MSU.EmptyModAddon).tostring(), (this.Keybinds == ::MSU.EmptyModAddon).tostring(), (this.ModSettings == ::MSU.EmptyModAddon).tostring());
+		return format("Mod %s, Versions %s\n", this.getID(), this.getVersionString())
```
LordMidas, line 69: semi-colon at the end

msu/msu_mod/setup_msu_mod.nut
```diff
@@ -0,0 +1 @@
+::MSU.Mod <- ::MSU.registerMod(::MSU.ID, ::MSU.Version, ::MSU.Name);
```
LordMidas, line 1: We should discuss if we want this wrapper for returning the mod object or retain the original method of instantiating `::MSU.Class.Mod`

msu/systems/debug/debug_mod_addon.nut
```diff
@@ -20,6 +20,16 @@
...
+	{
+		::MSU.System.Debug.setFlag(this.Mod.getID(), "default", true);
+	}
+
+	function disable()
```
LordMidas, line 28: As above

msu/systems/registry/load.nut
```diff
@@ -7,17 +7,18 @@ includeFile("registry_system.nut");
 local system = ::MSU.Class.RegistrySystem();
 
 ::MSU.System.Registry <- system;
-::MSU.Mods <- system.Mods;
+
+::MSU.registerMod <- function( _id, _version, _name = null )
```
LordMidas, line 11: As in a previous comment, we should discuss if we want this wrapper. Keep the discussion on that previous comment to keep it in one place.

msu/systems/registry/registry_system.nut
```diff
@@ -29,6 +29,21 @@
 		this.addMod(_mod);
 	}
 
+	function getMod( _modID )
+	{
+		if (_mod.getID() in this.Mods)
```
LordMidas, line 34: `_modID in this.Mods`

msu/systems/registry/registry_system.nut
```diff
@@ -29,6 +29,21 @@
...
+		{
+			::logError("Mod " + _mod.getID() + " not found in MSU! Did you forget to use ::MSU.registerMod()?");
```
LordMidas, line 36: `_modID` instead of `_mod.getID()`

msu/systems/registry/registry_system.nut
```diff
@@ -29,6 +29,21 @@
...
+			throw ::MSU.Exception.KeyNotFound(_mod.getID());
```
LordMidas, line 37: _modID

LordMidas (changes requested):

msu/systems/debug/debug_mod_addon.nut
TaroEld: It's supposed to be a very simplified and short helper function. People can read the documentation.

I don't get how having a separate enable and disable toggle helps in your example. If there are already settings for each flag, why do you need that on top to avoid log spam? Just tell the user to enable specific flags.
I don't like the idea of having to watch multiple things for the debugging to be enabled or disabled, not only the flag bool but then also your general enable and perhaps even fulldebug.

msu/msu_mod/setup_msu_mod.nut
Enduriel: I'm more in favor of the original system because it's simpler but I don't wanna die on that hill

msu/systems/debug/debug_mod_addon.nut
Enduriel: While I don't like that this change is being done in an unrelated PR, and initially I agreed with Midas, I think doing it this way is fine and honestly the separate boolean is a pretty niche system.
I think having it this way is fine and we just need to release.

msu/msu_mod/setup_msu_mod.nut
TaroEld: The wrapper has been removed

PR #97, opened 2022-04-15
Orderedmap tweaks

LordMidas:


msu/classes/ordered_map.nut
```diff
@@ -48,6 +49,22 @@
 		return ret;
 	}
 
+	function toTable()
+	{
+		return clone this.Table;
+	}
+
+	function addTable( _table, _overwrite = false )
```
Enduriel, line 57: I would default to true, since that's the standard behavior when using `<-` for a single key.

It's also more performant in situations where this doesn't come up (no `contains` check)

Enduriel (changes requested):
Other than that one comment though it all looks good.

PR #101, opened 2022-04-17
feat: add table utilities

LordMidas:


msu/utilities/table.nut
```diff
@@ -0,0 +1,79 @@
...
+		foreach (key, value in _table)
+		{
+			if (i == chosenIdx) return key;
+			i++;
+		}
+	}
+
+	function keys( _table )
+	{
+		local ret = array(_table.len(), null);
```
Enduriel, line 26: I believe the null is redundant

msu/utilities/table.nut
```diff
@@ -0,0 +1,79 @@
...
+		foreach (key, value in _table)
+		{
+			ret[i] = key;
+			i++;
+		}
+	}
+
+	function values( _table )
+	{
+		local ret = array(_table.len(), null);
```
Enduriel, line 37: As above

Enduriel (changes requested):
Other than those rly minor things looks good

PR #103, opened 2022-04-17
Change WeightedContainer to use a Table internally

LordMidas:
This helps to improve performance, while also allowing a `foreach (item, weight in container)` type iteration loop. It also gets rid of the ApplyIdx which was a potential pitfall for certain iteration cases.

### DIFF
--- a/msu/classes/weighted_container.nut
+++ b/msu/classes/weighted_container.nut
@@ -23,6 +23,15 @@
 		return ret;
 	}
 
+	function merge( _otherContainer )
+	{
+		::MSU.requireInstanceOf(::MSU.Class.WeightedContainer, _otherContainer);
+		foreach (item, weight in _otherContainer.Table)
+		{
+			this.add(item, weight);
+		}
+	}
+
 	function addArray( _array )
 	{
 		::MSU.requireArray(_array);
--- a/msu/classes/weighted_container.nut
+++ b/msu/classes/weighted_container.nut
@@ -252,7 +252,7 @@
 
 	function _cloned()
 	{
-		return ::MSU.Class.WeightedContainer(this.toArray(false));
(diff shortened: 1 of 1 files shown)


msu/classes/weighted_container.nut
```diff
@@ -1,25 +1,27 @@
 ::MSU.Class.WeightedContainer <- class
 {
 	Total = null;
-	Array = null;
-	ApplyIdx = null;
+	Table = null;
 	Forced = null;
+	ItemsArray = null;
```
Enduriel, line 6: Needs initialization

Enduriel (changes requested):
I don't think we should have *both* the array and the table. If we wanna switch this over and there are things a table means we can't do, that's fine, just remove those functions. I don't wanna duplicate this stuff in memory unless we have to (like with ordered map)

msu/classes/weighted_container.nut
LordMidas: > I don't think we should have _both_ the array and the table. If we wanna switch this over and there are things a table means we can't do, that's fine, just remove those functions. I don't wanna duplicate this stuff in memory unless we have to (like with ordered map)

The array is necessary for a foolproof foreach iteration (in order to allow a `foreach (item, weight in container)`. Otherwise I have to create an iteration array at the start of the foreach loop which reduces performance as I have to iterate over the entire table to first populate the array.

Ideally I wanted a `foreach (weight, item in container)` iteration but I haven't been able to do that. It might be doable if I create two arrays from the table at the start of the foreach. But that is really messy.

Alternatively we just disallow a foreach iteration?

Enduriel: Yeah, that's okay for me. If we keep apply/map/filter functions those are sufficient anyway.

PR #104, opened 2022-04-19
Modsettings improvements

TaroEld:
Improvements to mainly the system addon, the msu modsettings and some settingsElement things

### DIFF
--- a/msu/systems/mod_settings/settings_element.nut
+++ b/msu/systems/mod_settings/settings_element.nut
@@ -16,7 +16,8 @@
 		}
 		this.Name = _name == null ? _id : _name;
 		this.ID = _id;
-		this.Flags = this.new("scripts/tools/tag_collection");;
+		this.Flags = this.new("scripts/tools/tag_collection");
+		this.Flags.set("IsSetting", false);
 		this.Description = "";
 	}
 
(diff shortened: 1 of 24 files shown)


msu/systems/mod_settings/settings_element.nut
```diff
@@ -16,7 +16,7 @@
 		}
 		this.Name = _name == null ? _id : _name;
 		this.ID = _id;
-		this.Flags = [];
+		this.Flags = this.new("scripts/tools/tag_collection");;
```
Enduriel, line 19: I don't see any reason to use flag collection instead of a table for the changes you're making here.

Enduriel (changes requested):
For me it comes down to whether we want the modder working with objects or with strings/IDs. I'm very much in favor of objects since they're less error prone and you have stronger typing. With a string I feel like it's much easier to make a mistake. It's also often more verbose because of the additional "s you have to add.

Overall I'm in favor of having addPage return a page and the new function names, but I don't see a need for a separate divider function since it would be the only element with such a function.

The other problem with this sort of string based approach is it really limits opportunities for future expansion, with objects we can give them whatever properties and the modder can pass them around and it's easy to make backwards compatible changes, with function parameters it can be more difficult to for example add a required param after you already have an optional param for that function.

PR #111, opened 2022-04-25
fix: persistent data having broken prints and data

TaroEld:


Enduriel:
Merged in without a PR

PR #115, opened 2022-04-26
refactor: move semver related functions out of registry system and into their own utils nut

Enduriel:


LordMidas (approved):
I don't speak regex so looks ok to me if it looks ok to you and werks.

TaroEld (approved):

PR #129, opened 2022-05-08
Close #128 - Rework Affordability Preview so that current change is not required to be passed

### LINKED ISSUE #128: Rework Affordability Preview so that current change is not required to be passed

LordMidas:


msu/hooks/skills/skill_container.nut
```diff
@@ -189,16 +190,109 @@
...
 	{
+		this.PreviewProperty.clear();
 		foreach (skill in this.m.Skills)
 		{
 			skill.PreviewField.clear();
-			skill.PreviewProperty.clear();
 		}
 
-		this.callSkillsFunction("onAffordablePreview", [
+		this.doOnFunction("onAffordablePreview", [
```
Enduriel, line 199: Not sure why this was changed

LordMidas: Typo. Will fix.

Enduriel (approved):
After discussion, this needs a bit of a documentation improvement to make the functions more user accessible, but the general idea and code is probably as good as we can make it for now.

PR #132, opened 2022-05-12
feat: add function to get active state and use it in popup

TaroEld:


TaroEld:
Included in popup-improvements.

PR #134, opened 2022-05-13
Popup improvements and state management

TaroEld:
Improve popup code organization and add state getter that interacts with it.


scripts/config/msu/utils.nut
```diff
@@ -138,4 +139,15 @@
 		}
 		throw ::MSU.InvalidValue(_compareResult);
 	}
+
+	function getActiveState()
+	{
+		foreach (name, state in this.States)
+		{
+			if (state != null && state.isVisible() && name != "root_state")
```
Enduriel, line 147: Is that last one necessary since we hookChildren?

scripts/mods/msu/popup.nut
```diff
@@ -1,45 +1,58 @@
...
 	{
-		if (_forceQuit) this.m.ForceQuit = true;
-		if (this.m.JSHandle == null)
-		{
-			if (this.m.TextCache != "") this.m.TextCache += "<br>";
-			this.m.TextCache += _text;
-		}
-		else
+		this.m.JSHandle.asyncCall("addLine", _text);
+	}
```
Enduriel, line 10: The caching component has been completely removed here which is required for the popup to work

ui/mods/msu/css/popup.css
```diff
@@ -28,7 +28,7 @@
 	color: white;
 	font-family: monospace, arial;
 	font-size: 20;
-	background-color: rgba(0, 0, 0, 1.0);
+	/*background-color: rgba(0, 0, 0, 1.0);*/
```
Enduriel, line 31: If it's removed just delete it

Enduriel (changes requested):

scripts/config/msu/utils.nut
TaroEld: It is unfortunately

ui/mods/msu/css/popup.css
TaroEld: Was removed

PR #142, opened 2022-05-23
feat: add state management

TaroEld:
Add state management: keep track of them, provide functions to fetch states

### DIFF
--- a/msu/hooks/states/state.nut
+++ b/msu/hooks/states/state.nut
@@ -1,0 +1,8 @@
+::mods_hookChildren("states/state", function(o) {
+	local onInit = o.onInit;
+	o.onInit = function()
+	{
+		::MSU.Utils.States[this.ClassName] <- ::WeakTableRef(this);
+		return onInit();
+	}
+});
--- a/scripts/config/msu/utils.nut
+++ b/scripts/config/msu/utils.nut
@@ -7,6 +7,7 @@
 		Array = 4,
 		Table = 5
 	},
+	States = {},
 
 	function serialize( _object, _out )
 	{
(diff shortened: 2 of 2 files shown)


scripts/config/msu/utils.nut
```diff
@@ -138,4 +139,35 @@
...
+			if (state != null && state.isVisible() && name != "root_state")
+			{
+				return state;
+			}
+		}
+	}
+
+	function hasState(_id)
+	{
+		return (_id in this.States && this.States[_id] != null)
```
LordMidas, line 156: semi-colon

scripts/config/msu/utils.nut
```diff
@@ -138,4 +139,35 @@
...
 		throw ::MSU.InvalidValue(_compareResult);
 	}
+
+	function getActiveState()
+	{
+		foreach (name, state in this.States)
+		{
+			if (state != null && state.isVisible() && name != "root_state")
+			{
+				return state;
```
LordMidas, line 149: Will there always be at least one active state? If yes, then this is fine. Otherwise add handling for the case when no state is active/visible.

msu/hooks/states/state.nut
```diff
@@ -0,0 +1,8 @@
+::mods_hookChildren("states/state", function(o) {
+	local onInit = o.onInit;
+	o.onInit = function()
+	{
+		::MSU.Utils.States[this.ClassName] <- ::WeakTableRef(this);
```
LordMidas, line 5: Do state instances never go null after onInit?

LordMidas (changes requested):

PR #150, opened 2022-05-27
refactor: simplify movementspeed by removing a number of new paramete…

TaroEld:
Simplify movementspeed by removing a number of new parameters that were really superfluous and no longer overwriting onUpdate

### DIFF
--- a/msu/hooks/entity/world/party.nut
+++ b/msu/hooks/entity/world/party.nut
@@ -134,7 +134,7 @@
 		{
 			if (::MSU.Mod.Debug.isEnabled("movement") && ::Math.round(speedDelta) != ::Math.round(_speed))
 			{
-				this.testCompareMovementSpeeds(moddedSpeed, _speed)
+				this.__testCompareMovementSpeeds(moddedSpeed, _speed)
 			}
 			return move(_dest, speedDelta);
 		}
@@ -142,7 +142,7 @@
 		this.move = move;
 	}
 
-	o.testCompareMovementSpeeds <- function(_moddedSpeed, _speed)
+	o.__testCompareMovementSpeeds <- function(_moddedSpeed, _speed)
 	{
 		local name = this.m.IsPlayer ? "Player" : this.getName()
 		local vanillaMovementSpeed = 1.0
--- a/msu/hooks/entity/world/party.nut
+++ b/msu/hooks/entity/world/party.nut
@@ -1,6 +1,5 @@
 ::mods_hookExactClass("entity/world/party", function(o) {
-	o.m.VanillaBaseMovementSpeed <- o.m.BaseMovementSpeed;
-	o.m.BaseMovementSpeedMult <- 1.0;
+	// The final movement speed mult that is applied to the default value of 100
 	o.m.MovementSpeedMult <- 1.0;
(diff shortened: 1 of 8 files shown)


msu/hooks/entity/world/party.nut
```diff
@@ -332,43 +197,20 @@
...
 		}
-		this.resetBaseMovementSpeed();
 
+		::MSU.Mod.Debug.printError("Movement Speed for party " + this.getName() + " is NOT correct. Expected: " + speed + " , actual: " + _moddedSpeed, "movement")
+		logIfDifferent("GlobalMult", GlobalMult, this.getGlobalMovementSpeedMult());
+		logIfDifferent("slowDownPartyPerTroop", slowDownPartyPerTroop, this.getSlowdownPerUnitMovementSpeedMult());
+		logIfDifferent("NotPlayerMult", NotPlayerMult, this.getNotPlayerMovementSpeedMult());
+		logIfDifferent("RiverMult", RiverMult, this.getRiverMovementSpeedMult());
+		logIfDifferent("NighttimeMult", NighttimeMult, this.getNightTimeMovementSpeedMult());
+		logIfDifferent("TerrainTypeSpeedMult", TerrainTypeSpeedMult, this.getRoadMovementSpeedMult());
```
Enduriel, line 214: I assume this is a temp thing but you should probably be using popup, might be a system we wanna have incorporated

Enduriel (commented):
Besides the logIfDifferent thing, quite a fan of this so nice work

msu/hooks/entity/world/party.nut
TaroEld: Yeah I guess something like this could be used ina testing module

msu/hooks/entity/world/party.nut
```diff
@@ -136,148 +115,36 @@
...
-		::MSU.Mod.Debug.isEnabled("movement")
-		{
-			this.testCompareMovementSpeeds()
-		}
+		onUpdate();
+		this.move = move;
 	}
 
-	o.testCompareMovementSpeeds <- function()
+	o.testCompareMovementSpeeds <- function(_moddedSpeed, _speed)
```
LordMidas, line 145: Private functions should be prefixed with `__`

msu/hooks/entity/world/player_party.nut
```diff
@@ -70,4 +69,11 @@
 		}
 		return mult;
 	}
+
```
LordMidas, line 72: This is a comment regarding other stuff in this file that was present before this PR. All the cases where we are checking for `if ("xyz" in abc)` should be resolved in one of the following ways:
1. Use `if (::MSU.isIn("xyz", abc, true))` to account for inheritance and WeakTableRefs.
2. OR add the keys to the base classes with a default return of `1.0` so that we don't have to check for `in` every time. This might be more performant too.

LordMidas (changes requested):

msu/hooks/entity/world/player_party.nut
TaroEld: Did #2

msu/hooks/entity/world/party.nut
TaroEld: sure

PR #151, opened 2022-06-02
feat: add tooltip string coloring functions to MSU.String

Enduriel:


LordMidas (changes requested):
1. In my opinion these should be in MSU.Text namespace but I won't die on that hill.
2. Add a function called `::MSU.String.addNewColor( _name, _value )` that allows modders to create new colors in the ::Const.Color table. This is useful for example in PTR where I use a similar system to create colors for coding Passive, Active and One-Time Effect perk headings.
3. I would remove the colorBlue function as it is not used in any vanilla thing afaik. If it is, then keep it. Otherwise, only keep the colorGreen and colorRed. For other colors, modders can use the color( function.

scripts/config/msu/string.nut
```diff
@@ -27,4 +27,24 @@
 		}
 		return true;
 	}
+
+	function color( _string, _color )
```
LordMidas, line 31: the argument order should take _color first and then _string as it will look cleaner in code with longer strings.

LordMidas (changes requested):

TaroEld (approved):
lgtm

PR #155, opened 2022-06-07
feat: add multiply function

LordMidas:
This solves a particular thing in PTR where I have a list of multipliers for items in weighted containers and I want to multiply the weights. Currently, I have to do it via a custom helper function like following where I have to pass the container to the function. I think such a function being inside the container is helpful.
```squirrel
function applyMultipliers( _multipliersList, _treeContainer )
{
	foreach (multiplier in _multipliersList)
	{
		local tree = multiplier[1];
		if (_treeContainer.contains(tree)) _treeContainer.setWeight(tree, _treeContainer.getWeight(tree) * multiplier[0]);
	}
}
```

### DIFF
--- a/scripts/config/msu/classes/weighted_container.nut
+++ b/scripts/config/msu/classes/weighted_container.nut
@@ -127,6 +127,16 @@
 		this.updateWeight(_item, _weight);
 	}
 
+	function multiply( _multipliers )
+	{
+		// _multipliers is an array of len 2 arrays with idx 0 being item and idx 1 being multiplier
+
+		foreach (multiplier in _multipliers)
+		{
+			if (this.contains(multiplier[1])) this.setWeight(multiplier[1], this.getWeight(multiplier[1]) * multiplier[0]);
+		}
+	}
+
 	function apply( _function )
 	{
 		// _function (_item, _weight)


TaroEld:
Can't you use apply?

LordMidas:
> Can't you use apply?

Apply is highly inefficient in this case. An apply function would iterate over the multipliers array for every element in the container. E.g. like this:
```squirrel
local multipliers = someArray;
container.apply(function(_item, _weight) {
     foreach (multiplier in multipliers)
     {
          if (multiplier[1] == _item) return _weight * multiplier[0];
     }
});
```

Enduriel:
> > Can't you use apply?
> 
> `apply` is highly inefficient in this case. An apply function would iterate over the multipliers array for every element in the container. E.g. like this:
> 
> ```nut
> local multipliers = someArray;
> container.apply(function(_item, _weight) {
>      foreach (multiplier in multipliers)
>      {
>           if (multiplier[1] == _item) return _weight * multiplier[0];
>      }
> });
> ```
> 
> The proposed function in this PR avoids this problem by only iterating over the `multipliers` array once, and instead using `container.contains` multiple times which is much faster than an array iteration (for large containers) as the container is a table.

you should just be using a table for your multipliers, then there's no problem. I don't think this is a necessary function at all.

LordMidas:
> you should just be using a table for your multipliers, then there's no problem. I don't think this is a necessary function at all.

It is extremely inconvenient code-wise to use a table for multipliers because the `keys` can be anything e.g. objects. Compare, for example, the following two pieces of code:
```squirrel
// Array
multipliersArray = [
     [0, ::Const.Perks.SwordTree],
     [5, ::Const.Perks.CleaverTree],
     [-1, ::Const.Perks.PolearmTree],
]

// Table
multipliersTable = {};
multipliersTable[::Const.Perks.SwordTree] = 0;
multipliersTable[::Const.Perks.CleaverTree] = 5;
multipliersTable[::Const.Perks.PolearmTree] = -1;
```
The convenience and readability of the code both deteriorate with increasing size of the mulipliers table. Even simple strings e.g. perk ids cannot be stored in a simple syntax and have to be made as individual slots using the `[]` syntax. Therefore, I prefer the multipliers being an array.

**The second reason** for using an array is that the syntax is the same as the syntax used for the creation of the weighted container where we pass an array with len 2 arrays with idx 0 being weight and idx 1 being item. Therefore, it is more consistent to pass multipliers as a similarly formatted array.

Enduriel:
imo this is too niche to become a generic weighted container function and should remain as a helper function.

TaroEld:
I too think this is quite niche, and as you've learned recently you can use json style syntax in tables too so things like "perk.abc" isn't an issue.

LordMidas:
Alright. Closing without merge.

PR #156, opened 2022-06-08
feat: add helpful log errors to type checking functions

LordMidas:
I believe using `format` here has a chance of failure because as we saw earlier in the `getInQuotes` function, format doesn't seem to roll nicely with stuff that can't be converted to string? So we should instead use string concatenation here, right?

### DIFF
--- a/scripts/config/msu/type_checkers.nut
+++ b/scripts/config/msu/type_checkers.nut
@@ -39,6 +39,7 @@
 	{
 		if (typeof value != "instance" || !(value instanceof _class))
 		{
+			::logError(format("%s must be an instance of the class %s.", value, _class));
 			throw ::MSU.Exception.InvalidType(value);
 		}
 	}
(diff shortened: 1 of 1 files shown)


TaroEld:
> I believe using `format` here has a chance of failure because as we saw earlier in the `getInQuotes` function, format doesn't seem to roll nicely with stuff that can't be converted to string? So we should instead use string concatenation here, right?

I mean yeah, you can't just use %s there. Just use concatenation then.

Enduriel:
The concept seems good to me though

LordMidas:
Please review again. Also check syntax for `array.reduce`. Any way to pass class name instead of the class object in `requireInstance`?

TaroEld:
> Please review again. Also check syntax for `array.reduce`. Any way to pass class name instead of the class object in `requireInstance`?

class name will be passed if _tostring is defined. reduce syntax is good.

PR #163, opened 2022-07-01
feat: add onEquip and onUnquip events for skills

LordMidas:
The explicit Boolean check (ret == true) is because the vanilla unequip function can return null as well

### DIFF
--- a/msu/hooks/items/item_container.nut
+++ b/msu/hooks/items/item_container.nut
@@ -72,4 +72,20 @@
 		onNewRound();
 		this.m.ActionCost = ::Const.Tactical.Settings.SwitchItemAPCost;
 	}
+
+	local equip = o.equip;
+	o.equip = function( _item )
+	{
+		local ret = equip(_item);
+		if (ret == true && this.m.Actor != null) this.m.Actor.getSkills().onEquip(_item);
+		return ret;
+	}
+
+	local unequip = o.unequip;
+	o.unequip = function( _item )
+	{
+		local ret = unequip(_item);
+		if (ret == true && this.m.Actor != null) this.m.Actor.getSkills().onUnequip(_item);
+		return ret;
+	}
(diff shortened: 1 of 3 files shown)


TaroEld (approved):
LGTM

msu/hooks/items/item_container.nut
```diff
@@ -72,4 +72,20 @@
...
+		return ret;
+	}
+
+	local unequip = o.unequip;
+	o.unequip = function( _item )
+	{
+		local ret = unequip(_item);
+		if (ret == true && !::MSU.isNull(this.m.Actor) && this.m.Actor.isAlive()) this.m.Actor.getSkills().onUnequip(_item);
+		return ret;
+	}
```
Enduriel, line 90: onUnequip should be called before the item is unequipped (this is how items handle it). This requires us to redo some checks from item container first, but isn't actually that bad.

msu/hooks/items/item_container.nut
```diff
@@ -72,4 +72,20 @@
...
+		if (ret == true && !::MSU.isNull(this.m.Actor) && this.m.Actor.isAlive()) this.m.Actor.getSkills().onEquip(_item);
+		return ret;
+	}
+
+	local unequip = o.unequip;
+	o.unequip = function( _item )
+	{
+		local ret = unequip(_item);
+		if (ret == true && !::MSU.isNull(this.m.Actor) && this.m.Actor.isAlive()) this.m.Actor.getSkills().onUnequip(_item);
+		return ret;
```
Enduriel, line 89: onUnequip should be called before the item is unequipped (this is how items handle it). This requires us to redo some checks from item container first, but isn't actually that bad.

Enduriel (changes requested):

Enduriel (approved):

PR #168, opened 2022-07-25
feat: add onAnySkillUsed for items

LordMidas:


Enduriel (approved):

msu/hooks/skills/items/generic_item.nut
```diff
@@ -10,4 +10,16 @@
...
+
+	o.onAnySkillUsed <- function( _skill, _targetEntity, _properties )
+	{
+		if (!::MSU.isNull(this.m.Item) && !::MSU.isNull(this.m.Item.getContainer()))
+		{
+			this.m.Item.onAnySkillUsed(_skill, _targetEntity, _properties);
+		}
+		else
+		{
+			this.removeSelf();
```
TaroEld, line 22: Are you sure that this removeSelf() can't lead to issues?

TaroEld (commented):
Just the one query, otherwise lgtm

msu/hooks/skills/items/generic_item.nut
LordMidas: Yeah pretty sure that should be fine because all that function does is set `this.m.IsGarbage = true` in the skill which marks it for removal when the `collectGarbage` function of skill_container runs. It also excludes the skill from all sorts of iteration funtions e.g. `update` and `buildProperties` etc. so it's good.

TaroEld: Aight, good to go then

PR #169, opened 2022-07-25
fix: AdditionalAccuracy of items not being properly factored into the…

LordMidas:
… skill

### DIFF
--- a/msu/hooks/items/weapons/weapon.nut
+++ b/msu/hooks/items/weapons/weapon.nut
@@ -33,6 +33,18 @@
 		}
 	}
 
+	local addSkill = o.addSkill;
+	o.addSkill = function( _skill )
+	{
+		local ret = addSkill(_skill);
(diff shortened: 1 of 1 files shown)


msu/hooks/items/weapons/weapon.nut
```diff
@@ -33,6 +33,17 @@
...
+	local addSkill = o.addSkill;
+	o.addSkill = function( _skill )
+	{
+		addSkill(_skill);
+		if (::MSU.isIn("AdditionalAccuracy", _skill))
+		{
+			_skill.m.AdditionalAccuracy += this.m.AdditionalAccuracy;
+			_skill.setBaseValue("AdditionalAccuracy", _skill.m.AdditionalAccuracy);
+		}
+	}
```
Enduriel, line 45: return the result of wrapped addSkill, even if it doesn't return anything rn it's best practice

Enduriel (changes requested):
After that one tweak lgtm

PR #170, opened 2022-07-26
feat: add isEqual function

LordMidas:


Enduriel (commented):
~Taro~You made a cool suggestion about expanding this to not just work for WeakTableRefs, I thought it's worth considering but since our globals mostly deal with WeakTableRefs I'm not sure it's within scope.

LordMidas:
> ~Taro~You made a cool suggestion about expanding this to not just work for WeakTableRefs, I thought it's worth considering but since our globals mostly deal with WeakTableRefs I'm not sure it's within scope.

The way I've written it currently, it should work with everything, no?

Enduriel:
> > ~Taro~You made a cool suggestion about expanding this to not just work for WeakTableRefs, I thought it's worth considering but since our globals mostly deal with WeakTableRefs I'm not sure it's within scope.
> 
> The way I've written it currently, it should work with everything, no?

I thought you meant for it to handle some type casting, eg strings to ints and the like, but if not then we can go with this.

LordMidas:
> I thought you meant for it to handle some type casting, eg strings to ints and the like, but if not then we can go with this.

It is meant to be a replacement for `==` with handling for WeakTableRefs. Just like `isKindOf` etc. So I think we can go with the current implementation. It handles all data types just like `==` but with proper handling for WeakTableRefs of BB Objects.

Enduriel (approved):

TaroEld (approved):
lgtm

PR #171, opened 2022-07-26
DataType.Null

TaroEld:
Adds ::MSU.Utils.DataType.Null to deSer.

### DIFF
--- a/scripts/config/msu/utils.nut
+++ b/scripts/config/msu/utils.nut
@@ -5,7 +5,8 @@
 		Boolean = 2,
 		String = 3,
 		Array = 4,
-		Table = 5
+		Table = 5,
+		Null = 6
 	},
 	Timers = {}
 	States = {},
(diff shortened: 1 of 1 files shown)


Enduriel (approved):
When merging this in I'd squash the two null commits first but other than that all good from my side.

LordMidas:
Adding the Null data type should be a `feat` commit whereas changing the `::MSU.serialize` to `this.serialize` should be a separate `refactor` commit.

scripts/config/msu/utils.nut
```diff
@@ -60,12 +61,16 @@
 
 				case "array":
 					_out.writeU8(this.DataType.Array);
-					::MSU.serialize(element, _out);
+					this.serialize(element, _out);
```
LordMidas, line 64: Split the changing of `::MSU.` to `this` into a separate `refactor` commit to keep it separate from the `feat` of adding the Null data type.

LordMidas (changes requested):

scripts/config/msu/utils.nut
TaroEld: I'll just put that as a fix on dev

TaroEld:
This will get merged via other means

PR #175, opened 2022-07-27
feat: add an empty bb class that other classes can inherit from

LordMidas:
So a typical inheriting usage would look like:
```squirrel
this.my_new_bb_class <- ::inherit(::MSU.BBClass.Empty, {
    m = {
    },
    // my code
});
```

### DIFF
--- a/scripts/config/!msu.nut
+++ b/scripts/config/!msu.nut
@@ -5,6 +5,6 @@
 	VanillaID = "vanilla",
 	Class = {},
 	BBClass = {
-		Empty = "scripts/config/msu/classes/empty_bb_class"
+		Dummy = "scripts/config/msu/classes/dummy_bb_class"
 	}
 };
--- a/scripts/config/!msu.nut
+++ b/scripts/config/!msu.nut
@@ -3,5 +3,8 @@
 	ID = "mod_msu",
 	Name = "Modding Standards & Utilities (MSU)",
 	VanillaID = "vanilla",
-	Class = {}
+	Class = {},
+	BBClass = {
+		Empty = "scripts/config/msu/classes/empty_bb_class"
+	}
 };
--- a/scripts/config/!msu.nut
+++ b/scripts/config/!msu.nut
@@ -5,6 +5,6 @@
 	VanillaID = "vanilla",
 	Class = {},
 	BBClass = {
-		Dummy = "scripts/config/msu/bb_classes/dummy_bb_class"
+		Empty = "sripts/mods/msu/bb_classes/empty_bb_class"
 	}
 };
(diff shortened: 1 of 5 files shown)


scripts/config/msu/classes/empty_bb_class.nut
```diff
@@ -0,0 +1,6 @@
+this.empty_bb_class <- {
+	m = {},
+	function create()
+	{
+	}
+}
```
Enduriel, line 6: I think this should be in a different folder, maybe bbclasses. I don't think it should be grouped together with Squirrel classes

TaroEld:
Call it dummy_bb_class imp

LordMidas:
@Enduriel , @TaroEld Changes made. Please re-review.

scripts/config/!msu.nut
```diff
@@ -3,5 +3,8 @@
 	ID = "mod_msu",
 	Name = "Modding Standards & Utilities (MSU)",
 	VanillaID = "vanilla",
-	Class = {}
+	Class = {},
+	BBClass = {
+		Dummy = "scripts/config/msu/classes/dummy_bb_class"
```
Enduriel, line 8: Path is bad now

TaroEld, line 8: Perhaps just put it in the Class table with DummyBBClass as the name?

LordMidas: No, I think a separate BBClass namespace is much better - as both classes are instantiated in different ways e.g. `local myNewSQClass = ::MSU.Class.WeightedContainer()` vs `local myNewBBClass = ::new(::MSU.BBClass.SomeClass)`.

Enduriel (approved):
lgtm now

LordMidas:
@Enduriel The only way the inheriting and creating while also allowing hooking the dummy class by using the `::MSU.BBClass.Dummy` can work is if we implement #187.

LordMidas:
Review again please. I've moved it to `scripts/mods/msu/bb_classes` so that it can be both inherited from and hooked via `::MSU.BBClass.Empty`.

Enduriel (approved):

TaroEld (approved):

LordMidas:
@Enduriel @TaroEld So, I suggest we remove the `create` function from this "empty" class because apparently that is not required to instantiate any class and several vanilla BB base classes e.g. `skill.nut` do not have a `create` function. If we ever get a bug report that something isn't working because `create` is missing from this empty class, we can then add it.

Enduriel:
Oh wow, thanks for pointing that out, no I'm not okay with the folder structure. The BB_Classes folder is redundant, if it's in scripts and not in !mods_preload or config, it's a bb class. This doesn't need to be stated again.

Enduriel:
I think leaving it in is fine, I don't care enough about it to die on this hill though, I don't think it makes a difference.

LordMidas:
@Enduriel final look please

Enduriel (approved):

PR #176, opened 2022-07-28
folder restructure

LordMidas:
Key changes:
1. Move `msu/config` to `msu/hooks/config`.
2. Move `utils` and `classes` out of `scripts/config` and into `msu`.
3. Load `msu/utils` and `msu/classes` during `scripts/config/!msu.nut` for loading ahead of mods_queue.
4. Move `::includeFiles` etc. functions to the `::MSU.` namespace. These should remain undocumented imo as they aren't for others to use.


Enduriel (approved):
I mean sure

TaroEld (approved):
lgtm assuming you tried it and there isn't some whacky load order issue

scripts/config/!msu.nut
```diff
@@ -5,3 +5,24 @@
 	VanillaID = "vanilla",
 	Class = {}
 };
+
+::includeLoad <- function( _prefix, _folder )
```
LordMidas, line 9: Perhaps these functions should be in `::MSU.` namespace?

scripts/config/!msu.nut
```diff
@@ -5,3 +5,24 @@
...
+	foreach (file in _files)
+	{
+		if (_includeLoad || split(file, "/").pop() != "load.nut")
+		{
+			::include(file);
+		}
+	}
+}
+
+::includeFiles(::IO.enumerateFiles("scripts/config/msu"));
```
LordMidas, line 28: As we have `!msu.nut` in the `scripts/config` folder, perhaps it is best to move the Classes etc. that we moved to that folder for the idea of "loading them before hooks" back to the `/msu` folder and instead use this method to load them early i.e. during config load?

Enduriel (approved):
Sure, I actually like it better this way as well

Enduriel (changes requested):
After the moves of ::includeFile, ::includeFiles and ::includeLoad, I think this has moved outside the scope of the initial PR, and I'd want some more discussion on whether these moves are really for the best.

LordMidas:
This PR will be merged when all other PRs for this release have been merged - due to the folder structure change. Any other files/changes that were added in other PRs that need to be moved will be moved and added to this PR at that time.

LordMidas:
Reminder for self: Also move `msu/config` to `msu/hooks/config`

Enduriel (approved):

TaroEld (approved):

PR #177, opened 2022-07-31
Validate keybinds

TaroEld:
Validates keybinds before creating them. I readded the 'reverse' keymaps for fast checks.


TaroEld:
bad base

PR #179, opened 2022-08-01
feat: add more utility functions to get all settings

TaroEld:
Mostly I wanted it for the reset button and since I'm a nice guy I made them a separate PR

### DIFF
--- a/msu/systems/mod_settings/settings_page.nut
+++ b/msu/systems/mod_settings/settings_page.nut
@@ -109,7 +109,7 @@
 		return this.Settings;
 	}
 
-	function getSettingsAsArray()
+	function getAllSettingsAsArray()
 	{
 		return this.getSettings().values();
 	}
--- a/msu/systems/mod_settings/settings_page.nut
+++ b/msu/systems/mod_settings/settings_page.nut
@@ -109,6 +109,11 @@
 		return this.Settings;
 	}
 
+	function getSettingsAsArray()
+	{
+		return this.getSettings().values();
+	}
+
 	function get( _settingID )
 	{
 		return this.Settings[_settingID];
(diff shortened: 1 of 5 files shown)


msu/systems/mod_settings/settings_page.nut
```diff
@@ -109,6 +109,11 @@
 		return this.Settings;
 	}
 
+	function getSettingsAsArray()
```
LordMidas, line 112: Same comment as on the other PR. Why not call this `getAllSettings()` too?

TaroEld: The reasoning was that getSettings() already returns all the settings, but in a different format. So it's kind of unclear what the difference is just from the function name. But I'm open to changing it.

LordMidas: I see. So, without this function one would have to call `getSettings()` to get `this.Settings` which is an instance of OrderedMap and then manually extract an array of its values which are the settings objects? But why would one want to do that? I mean the OrderedMap can be iterated on with foreach and has apply/filter/map functions. What is the advantage of returning the array?

TaroEld: In my experience, gettinf an OrderedMap when you're not expecting it is bad mojo. In the reset PR I ran into the issue where I wanted to .extend() another array with it, which naturally didn't work. The second function makes it more clear what you're getting with getSettings(), ie, not a simple array.

LordMidas: In that case perhaps a reasonable name could be `getAllSettingsAsArray`?

TaroEld: Sure, I don't mind. Can also leave it away entirely and make it clear through the wiki what you're getting, but I'm not sure.

Enduriel: imo they should all be called `getAllSettingsAsArray` because that's not the 'default' container for settings and should therefore be marked as different somehow.

TaroEld: Sure, works for me.

Enduriel (approved):
lgtm

PR #181, opened 2022-08-02
Pay for action fixes

LordMidas:


msu/hooks/ai/tactical/behaviors/ai_throw_bomb.nut
```diff
@@ -0,0 +1,13 @@
...
+	local onExecute = o.onExecute;
+	o.onExecute = function( _entity )
+	{
+		local itemsBefore = _entity.getItems().getAllItems();
+		local ret = onExecute(_entity);
+		if (ret && this.m.Skill == null)
+		{
+			this.getItems().payForAction(_entity.getItems().getAllItems().filter(@(idx, item) itemsBefore.find(item) == null));
+		}
+		return ret;
```
Enduriel, line 11: why can't the other 2 hooks use this method?

LordMidas: Because in those cases the total items don't change (aka the entity is not adding/removing items to their slots, only shifting their positions). So I'd instead have to go through each slot and see if the item in it changed instead of just comparing the total items before and total items after. Should I implement going through each slot and checking if it changed? Maybe it is more robust like that? (It will still need special handling for the Disarmed thing but that's ok).

TaroEld: Yes, going through a few slots is nothing.

LordMidas: Ok I will do it like that.

Enduriel: Yeah, this would be much more robust, ideally we should have a single standardized wrapper for this situation.

msu/hooks/entity/tactical/actor.nut
```diff
@@ -68,6 +68,27 @@
 		return ::Math.floor(getActionPointsMax());
 	}
 
+	// This function is called during `onExecute` of the `ai_pickup_weapon` behavior
+	// This hook is necessary to ensure that the correct items array is passed to the `payForAction` function
+	// TODO - needs discussion because in vanilla even if the AI picks up a weapon AND a shield, they only pay the cost ONCE i.e. 4 AP
+	// However with this change now they will be able to pick up a weapon with 0 Cost if they have Quick Hands - this conflicts with vanilla behavior
+	// We could leave this hook out completely but that will leave a loop hole in MSU implementation regarding skills which may affect AP cost of picking
+	// up items from the ground
+	local pickupMeleeWeaponAndShield = o.pickupMeleeWeaponAndShield;
```
LordMidas, line 77: @TaroEld and @Enduriel Need your input on this "TODO" discussion point.

Enduriel: How does this work for players? Afaik players can't do this so this is a bug in vanilla that we would be fixing.

msu/hooks/items/item_container.nut
```diff
@@ -1,5 +1,6 @@
 ::mods_hookNewObject("items/item_container", function(o) {
 	o.m.ActionSkill <- null;
+	o.m.IsMSUHandledItemAction <- false;
```
Enduriel, line 3: I would call this `MSU_IsHandledItemAction`

Enduriel (changes requested):
Instead of copy-pasting the same hook 3 times, I think it would be better to create a public wrapper function that modders could potentially also use.

msu/hooks/entity/tactical/actor.nut
LordMidas: > How does this work for players? Afaik players can't do this so this is a bug in vanilla that we would be fixing.

This function is only for AI. And I think this isn't a bug but intended behavior in vanilla i.e. an AI entity is supposed to be able to pick up the gear (both weapon and shield) with a single item action.

msu/hooks/ai/tactical/behaviors/ai_switchto_ranged.nut
```diff
@@ -0,0 +1,31 @@
...
+	local onExecute = o.onExecute;
+	o.onExecute = function( _entity )
+	{
+		_entity.getItems().m.IsMSUHandledItemAction = true;
+
+		local itemsBefore = {};
+		for (local i = 0; i < ::Const.ItemSlot.COUNT; i++)
+		{
+			itemsBefore[i] = _entity.getItems().getAllItemsAtSlot(i);
+		}
```
TaroEld, line 11: That should probably be a generic 'getAllItemsInEverySlot' function

msu/hooks/ai/tactical/behaviors/ai_switchto_ranged.nut
```diff
@@ -0,0 +1,31 @@
...
+		if (ret)
+		{
+
+			local items = [];
+			for (local i = 0; i < ::Const.ItemSlot.COUNT; i++)
+			{
+				local itemsNowInSlot = _entity.getItems().getAllItemsAtSlot(i);
+				items.extend(itemsBefore[i].filter(@(idx, item) itemsNowInSlot.find(item) == null));
+				items.extend(itemsNowInSlot.filter(@(idx, item) itemsBefore[i].find(item) == null));
+			}
```
TaroEld, line 23: Which you can then also call here and subsequently do the filtering/comparing

msu/hooks/items/item_container.nut
```diff
@@ -9,6 +10,8 @@
 
 	o.getActionCost = function( _items )
 	{
+		if (this.m.IsMSUHandledItemAction) return 0;
```
TaroEld, line 13: Doesn't this mess with tooltips?

msu/hooks/entity/tactical/actor.nut
Enduriel: I don't like that ngl, I think we should maybe report this as a bug and see what overhype says?

LordMidas: This function is only used for resurrecting skeletons and zombies in vanilla. Therefore I say that it is intended behavior (the resurrecting zombie/skeleton uses this function to pick up the gear on their tile). So, I suppose this function should be removed from our handling anyway?
```
Searching 4858 files for ".pickupMeleeWeaponAndShield("

D:\BB-Modding\vanilla\scripts\ai\tactical\behaviors\ai_pickup_weapon.nut:
   61  		}
   62  
   63: 		_entity.pickupMeleeWeaponAndShield(_entity.getTile());
   64  		_entity.getItems().payForAction([]);
   65  		return true;

D:\BB-Modding\vanilla\scripts\entity\tactical\skeleton.nut:
  367  		this.actor.onResurrected(_info);
  368  		this.m.IsResurrected = true;
  369: 		this.pickupMeleeWeaponAndShield(this.getTile());
  370  		this.getSkills().update();
  371  		this.m.XP /= 4;

D:\BB-Modding\vanilla\scripts\entity\tactical\enemies\zombie.nut:
  496  		this.actor.onResurrected(_info);
  497  		this.m.IsResurrected = true;
  498: 		this.pickupMeleeWeaponAndShield(this.getTile());
  499  		this.getSkills().update();
  500  		this.m.XP /= 4;

3 matches across 3 files
```

msu/hooks/ai/tactical/behaviors/ai_switchto_ranged.nut
LordMidas: There already exists in vanilla `item_container` a function called `getAllItems` which returns an array of all the items at all slots. However, in this particular case I want the items to be pulled into a table where each `key` is a `slot` and the values of that key are the items in that slot.

msu/hooks/items/item_container.nut
LordMidas: No, because these are only for AI. Also, returning 0 for a player doesn't mess any tooltip, it just mentions the cost of the item swap to be 0 AP.

msu/hooks/ai/tactical/behaviors/ai_throw_bomb.nut
LordMidas: Unresolving this. I've thought about it a bit more and I think having a wrapper for this is unnecessary. If a modder is adding something that swaps items on an AI entity, he can just pass those items to the `payForAction` function in place. The only reason we need to do this comparing thing here is because we want to hook a vanilla function and want to be safe against future changes to that function. So it is a one-off application of this method, which in my opinion has no use for modders.

msu/hooks/entity/tactical/actor.nut
Enduriel: Yeah actually that actually explains a lot. I guess just make sure if we get passed nothing we do nothing...

msu/hooks/ai/tactical/behaviors/ai_throw_bomb.nut
Enduriel: After reading your other comment I agree.

msu/hooks/items/item_container.nut
TaroEld: What do you mean? That function is used in tooltips. And showing 0 cost when it's not 0 cost is bad?

Enduriel: He's saying that IsMSUHandledItemAction would only be true for AI skills @TaroEld (that's the first part anyway)

TaroEld: I see

msu/hooks/entity/tactical/actor.nut
LordMidas: No it isn't as simple as that. Vanilla passes nothing because they want the entity to spend 4 AP (the default item action cost). So, that's how I have it set up right now:
```squirrel
o.getActionCost = function( _items )
{
	if (this.m.IsMSUHandledItemAction) return 0;

	this.m.ActionSkill = null;

	local info = this.getActor().getSkills().getItemActionCost(_items);

	info.sort(@(info1, info2) info1.Skill.getItemActionOrder() <=> info2.Skill.getItemActionOrder());

	local cost = ::Const.Tactical.Settings.SwitchItemAPCost;

	foreach (entry in info)
	{
		if (entry.Cost < cost)
		{
			cost = entry.Cost;
			this.m.ActionSkill = entry.Skill;
		}
	}

	return cost;
}

o.payForAction = function ( _items )
{
	local actionCost = this.getActionCost(_items);
	this.m.Actor.setActionPoints(::Math.max(0, this.m.Actor.getActionPoints() - actionCost));
	if (_items.len() != 0) this.m.Actor.getSkills().onPayForItemAction(this.m.ActionSkill, _items);
	this.m.ActionSkill = null;
}
```
So, basically `getActionCost` returns the default cost if `_items` is an empty array. However, if it is an MSU Handled Item Action (which is a switch we ONLY use for handling AI related vanilla item actions), then it returns 0.

The main question that I am a bit confused about and would like input from you and @TaroEld here is whether we call the `onPayForItemAction` event for skills if `_items` is empty or not. Currently, I've set it to not call that, because `_items` is only empty during AI related actions and calling that event can trigger downstream changes in skills which have that event and do something based on that.

Enduriel (changes requested):
I don't know how we haven't thought of this before but add a function that creates these three hooks when passed the file to hook and the function, something like
```squirrel
function fixPayForAction( _scriptPath, _functionName )
{
	::mods_hookExactClass(_scriptPath, function (o)
	{
		local func = o[_functionName];
		o[_functionName] = function( _entity )
		{
			// do stuff before
			local ret = func(_entity);
			// do stuff after
			return ret;
		}
	}
}
```
and then we could just do
```
fixPayForAction("ai/tactical/behaviors/ai_switchto_melee")
```
in each of the files (so that we are still clear on what files we're hooking)

LordMidas:
> I don't know how we haven't thought of this before but add a function that creates these three hooks when passed the file to hook and the function, something like
> ```squirrel
> function fixPayForAction( _scriptPath, _functionName )
> {
> 	::mods_hookExactClass(_scriptPath, function (o)
> 	{
> 		local func = o[_functionName];
> 		o[_functionName] = function( _entity )
> 		{
> 			// do stuff before
> 			local ret = func(_entity);
> 			// do stuff after
> 			return ret;
> 		}
> 	}
> }
> ```
> and then we could just do
> ```
> fixPayForAction("ai/tactical/behaviors/ai_switchto_melee")
> ```
> in each of the files (so that we are still clear on what files we're hooking)

I did think of this and didn't do it because the switch to melee file needs a special case treatment. We could do a loop hook but then we'd need a closure with script name to do the special case.

Enduriel (approved):

LordMidas:
@Enduriel I think we don't need to add the `isSlotEmptyOrBlocked` function. Shall I remove it from this PR?

Enduriel:
> @Enduriel I think we don't need to add the `isSlotEmptyOrBlocked` function. Shall I remove it from this PR?

yes

TaroEld (approved):

PR #184, opened 2022-08-05
fix: closure issue causing all events to use parameters from the last…

LordMidas:
… iteration of the foreach loop


TaroEld (approved):

Enduriel (approved):
lgtm

PR #187, opened 2022-08-14
feat: add a hook on mod hooks to allow passing names that start with …

LordMidas:
…`scripts/`

### DIFF
--- a/msu/squirrel_hooks/mod_hooks.nut
+++ b/msu/squirrel_hooks/mod_hooks.nut
@@ -17,6 +17,13 @@ local function getRegisteredModRef( _modID )
 }
 local lastRegistered = null;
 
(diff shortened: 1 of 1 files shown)


LordMidas:
> I mean the issue is obviously that this isn't 100% backwards compatible, afaik this shouldn't break any mods that are out there, but in theory it could. Maybe instead log an error into the log when this happens to warn modders without breaking compatibility?

I don't see how this isn't 100% backwards compatible?

TaroEld:
I don't really see the point, this just muddies things. People will see MSU-based mods hooking with scripts/, try it in their non-MSU mods and it won't work. 
A better change would be an eventual change to modhooks that allows hooking files not in scripts.

PR #188, opened 2022-08-14
Colorize values (Close #162)

### LINKED ISSUE #162: [REQUEST] Helper functions to create tooltips for positve/negative tooltips
Often times when writing future-proof skills and effects for Battle Brothers they also need to show positive or negative values in tooltips for them.
Currently this is either done by 
- a lengthy if/else statement depending on if the value is positive or negative
or 
- the tooltip is hard-coded which means it needs to be adjusted every time that the skill changes

Three helping functions that work something like this:
image

With the help of those we can now create versatile but compact tooltips
image

LordMidas:
Close Issue #162

### DIFF
--- a/scripts/config/msu/text.nut
+++ b/scripts/config/msu/text.nut
@@ -13,4 +13,32 @@
 	{
 		return this.color(::Const.UI.Color.NegativeValue, _string);
 	}
+
+	function colorizeValue( _value, _addSign = true, _compareTo = 0, _invertColor = false, _addPercent = false )
+	{
+		if (_value < _compareTo)
+		{
+			if (!_addSign && _value < 0) _value *= -1;
+			if (_addPercent) _value = _value + "%";
+			return _invertColor ? this.colorGreen(_value) : this.colorRed(_value);
+		}
+
+		if (_value > _compareTo)
+		{
+			if (_addSign && _value > 0) _value = "+" + _value;
+			if (_addPercent) _value = _value + "%";			
+			return _invertColor ? this.colorRed(_value) : this.colorGreen(_value);
+		}
+
+		if (_value == _compareTo)
+		{
+			if (_addPercent) _value = _value + "%";
+			return _value;
+		}
+	}
+
+	function colorizePercentage( _value, _addSign = true, _compareTo = 0, _invertColor = false )
+	{
+		return this.colorizeValue(_value, _compareTo, _addSign, _invertColor, true);
+	}
 }
--- a/scripts/config/msu/text.nut
+++ b/scripts/config/msu/text.nut
@@ -1,7 +1,7 @@
 ::MSU.Text <- {
 	function color( _color, _string )
 	{
-		return format("[color=%s]%s[/color]", _color, _string);
+		return "[color=" + _color + "]" + _string + "[/color]";
 	}
 
 	function colorGreen( _string )
--- a/scripts/config/msu/text.nut
+++ b/scripts/config/msu/text.nut
@@ -1,7 +1,7 @@
 ::MSU.Text <- {
 	function color( _color, _string )
 	{
-		return "[color=" + _color + "]" + _string + "[/color]";
+		return ::Const.UI.getColorized(_string, _color);
 	}
 
 	function colorGreen( _string )


scripts/config/msu/text.nut
```diff
@@ -13,4 +13,23 @@
...
+
+		if (_value == 0) return this[_invert ? "colorRed" : "colorGreen"](_value);
+	}
+
+	function getIncreaseDecrease( _value, _invert = false )
+	{
+		return (_value >= 0 || _invert) ? "increase" : "decrease";
+	}
+
+	function getIncreaseReduce( _value, _invert = false )
```
LordMidas, line 31: I've added separate functions for `decrease` and `reduce` because I think both words fit differently in a sentence based on the context. However, we could remove one of these functions. What do you say?

scripts/config/msu/text.nut
```diff
@@ -13,4 +13,23 @@
...
 	}
+
+	function colorizeValue( _value, _invert = false, _addSign = true )
+	{
+		if (_value < 0) return this[_invert ? "colorGreen" : "colorRed"](_value);
+
+		if (_value > 0) return (_addSign ? "+" : "") + this[_invert ? "colorRed" : "colorGreen"](_value);
+
+		if (_value == 0) return this[_invert ? "colorRed" : "colorGreen"](_value);
+	}
```
Enduriel, line 24: We should not be referring to functions using strings like this, no matter how convenient it is, it causes issues further down the line with it being unclear when this stuff is used as a function when searching.

scripts/config/msu/text.nut
```diff
@@ -13,4 +13,23 @@
...
 		return this.color(::Const.UI.Color.NegativeValue, _string);
 	}
+
+	function colorizeValue( _value, _invert = false, _addSign = true )
+	{
+		if (_value < 0) return this[_invert ? "colorGreen" : "colorRed"](_value);
+
+		if (_value > 0) return (_addSign ? "+" : "") + this[_invert ? "colorRed" : "colorGreen"](_value);
+
+		if (_value == 0) return this[_invert ? "colorRed" : "colorGreen"](_value);
```
Enduriel, line 23: I would just make it optionally not color at all in this case (another parameter lol)

scripts/config/msu/text.nut
```diff
@@ -13,4 +13,23 @@
 	{
 		return this.color(::Const.UI.Color.NegativeValue, _string);
 	}
+
+	function colorizeValue( _value, _invert = false, _addSign = true )
+	{
+		if (_value < 0) return this[_invert ? "colorGreen" : "colorRed"](_value);
```
Enduriel, line 19: Missing - from `_addSign`

scripts/config/msu/text.nut
Enduriel: I think increase & decrease is fine unless you have a specific example where you'd wanna use increase/reduce in which case leave both.

Enduriel (changes requested):

scripts/config/msu/text.nut
LordMidas: > specific example where you'd wanna use increase/reduce
Reduced example: Damage to Hitpoints is **reduced** by x%. In this case "decreased" doesn't sound as good.

Honestly I think "reduce" will be used much more than "decrease".

scripts/config/msu/text.nut
LordMidas: Negative doesn't need to be added because negative numbers already print with a negative sign in strings.

scripts/config/msu/text.nut
LordMidas: Ok, I will change it.

scripts/config/msu/text.nut
LordMidas: I don't see the point of it. Could you elaborate on why it is useful?

scripts/config/msu/text.nut
Enduriel: I'm an idiot, nice

Enduriel: Actually in that case, I'd *remove* the sign if `_addSign` is false. This would come in useful if you were to say something like "increases Thing by x" and "decreases Thing by x" rather than "increases Thing by -/+ x"

scripts/config/msu/text.nut
Enduriel: Maybe a different function for this kinda thing would be better, allow the modder to pass their increase/decrease equivalent values?

scripts/config/msu/text.nut
Enduriel: If you're showing a range of stuff you may want 0 to be neutral (and therefore black)

LordMidas: In that case you wouldn't use this function on the first part of the range. Because there is no way to turn the colorization on/off based on the value.

scripts/config/msu/text.nut
LordMidas: Sounds good, I will try to write something up for that.

scripts/config/msu/text.nut
```diff
@@ -1,7 +1,7 @@
 ::MSU.Text <- {
 	function color( _color, _string )
 	{
-		return format("[color=%s]%s[/color]", _color, _string);
+		return "[color=" + _color + "]" + _string + "[/color]";
 	}
 
 	function colorGreen( _string )
```
TaroEld, line 7: looks like bad spacing here or is that just github

scripts/config/msu/text.nut
```diff
@@ -13,4 +13,39 @@
...
+
+	function getChangeVerb( _value, _compareTo = 0, _invert = false, _upVerb = "increase", _downVerb = "decrease" )
+	{
+		if (_value >= _compareTo || _invert) return _upVerb : _downVerb;
+	}
+
+	function getIncreaseReduce( _value, _compareTo = 0, _invert = false )
+	{
+		return this.getChangeVerb(_value, _compareTo, _invert, "increase", "reduce");
+	}
```
TaroEld, line 50: Are these actually useful? Just using a ternary in place seems like the same effort and you don't need to remember the functions.

scripts/config/msu/text.nut
TaroEld: Below the lines indicated here

Enduriel: It looks fine to me?

scripts/config/msu/text.nut
Enduriel: That's actually a good point, maybe this should color?

scripts/config/msu/text.nut
TaroEld: then it is probably github

scripts/config/msu/text.nut
LordMidas: > Are these actually useful? Just using a ternary in place seems like the same effort and you don't need to remember the functions.

Wdym?

LordMidas: > That's actually a good point, maybe this should color?

You mean color the word?

Enduriel: Yes

TaroEld: > Wdym?

It is already simple to get such terms with a ternary. The simple case is:
val > 0 ? "increase" : "decrease" ; getChangeVerb(val)
But then you have cases where it's other than 0
val > 1 ? "increase" : "decrease" ; getChangeVerb(val, 1)
Or inverted
val < 0 ? "increase" : "decrease" ; getChangeVerb(val, 0, true)
Or different verbs
val < 0 ? "increase" : "reduce" ; getIncreaseReduce(val, 0, true)
Or the temp doesn't fit
val < 0 ? "increases" : "decreases" ; ???
Or different conditions
val != null && val = 0 ? "increases" : "decreases" ; ???

Basically I think it's barely worth, it's going to be more effort to remember what the function is called, and if you need a bunch of those you just make a local function

LordMidas: I get your point now but just want to point out that the following is unnecessary:
> Or the temp doesn't fit
val > 0 ? "increases" : "decreases" ; getChangeVerb(val, 0, false, "increases", "decreases")

Because you can simply append `s` or `d` at the end of the returned verb when typing your tooltip to convert it into the tense you want.

So, in a string it will look something like:
```
"This is my tooltip string with an " + (val >= 0 ? "increase" : "decrease") + " of something something which leads to something something"
```

vs
```
"This is my tooltip string with an " + ::MSU.Text.getChangeVerb(val) + " of something something which leads to something something"
```

So, yeah I think the ternary is fine in this case.

LordMidas:
@Enduriel , @TaroEld Can you please suggest a better name than `colorizeValueAndEqual`?

Enduriel:
Is there anything more that still needs discussing here?

LordMidas:
> Is there anything more that still needs discussing here?

Not from my side. From my side this is good to go, if you guys agree.

Enduriel (approved):

scripts/config/msu/text.nut
```diff
@@ -13,4 +13,32 @@
 	{
 		return this.color(::Const.UI.Color.NegativeValue, _string);
 	}
+
+	function colorizeValue( _value, _compareTo = 0, _addSign = true, _invertColor = false, _addPercent = false )
+	{
+		if (_value < _compareTo)
+		{
+			if (!_addSign) _value *= -1;
+			if (_addPercent) _value = _value + "%";
```
TaroEld, line 22: _addPercent seems repeated four times, just do it at the beginning

TaroEld (approved):
That change then good

scripts/config/msu/text.nut
TaroEld: Ah I see you can't do that due to the negative case. Might want to define a return string and manipulate that instead, but doesn't matter much,

LordMidas:
@Enduriel and @TaroEld Should `_addSign` be the second argument instead of third? I've kept it as a third in the original version because `_compareTo` made more "logical" sense to be second arg. However, in terms of use, `_compareTo` will almost always be 0. Therefore, it makes it more clean if `_addSign` is second because that is the parameter that is most often changed.

I've tried using this function already in a lot of strings the parameter I ended up changing often was `_addSign` and in the current implementation this requires me writing `colorizePercentage(myValue, 0, false)` which is annoying compared to `colorizePercentage(myValue, false)`.

PR #189, opened 2022-08-19
refactor: remove injuries handler effect and instead move the code directly to skill container onBeforeTargetHit

LordMidas:
This ensures that the code always runs before the event is called for any skills. Even though we had SkillOrder of ::Const.SkillOrder.First in the injuries handler effect, it is not guaranteed to run before any other skill with the same order. This also removes the need to add an extra skill to actors, and removes the hook on onInit of actor.


Enduriel (approved):
looks much better to me

PR #191, opened 2022-08-19
feat: adds an enum class

Enduriel:
The goal here is to simplify the creation of an extensible enum.
Squirrel Enums are not extensible by design, and BB Enums don't have any of the convenience functions that would allow modders to easily add their own values (or any of the safety checks to make sure values are ordered).

Imo we might wanna add a second Enum Class that handles enums with arbitrary values that can be defined by the modder. I'd initially made it be just a toggle set during the instantiation, but that overcomplicated some of the aspects of the class. It also shouldn't be a child/parent class of this, but maybe they can both inherit from the same abstract ancestor. I think we can leave it for now and get to it at a later point.

### DIFF
--- a/scripts/config/msu/classes/enum.nut
+++ b/scripts/config/msu/classes/enum.nut
@@ -2,7 +2,6 @@
 {
 	__Keys = null;
 	__ValueKeyArray = null;
-	__NextICache = null;
 
 	constructor( _array = null )
 	{
--- a/scripts/config/msu/classes/enum.nut
+++ b/scripts/config/msu/classes/enum.nut
@@ -41,6 +41,12 @@
 		return this.__Keys.len();
 	}
 
+	function contains( _value )
+	{
+		::MSU.requireInt(_value);
+		return _value >= 0 && _value < this.len();
+	}
+
 	function _get( _key )
 	{
 		if (_key in this.__Keys) return this.__Keys[_key];
--- a/scripts/config/msu/classes/enum.nut
+++ b/scripts/config/msu/classes/enum.nut
@@ -1,29 +1,30 @@
 ::MSU.Class.Enum <- class
 {
-	_Keys = null;
-	_ValueKeyArray = null;
+	__Keys = null;
+	__ValueKeyArray = null;
+	__NextICache = null;
 
 	constructor( _array = null )
 	{
-		this._Keys = {};
-		this._ValueKeyArray = [];
+		this.__Keys = {};
+		this.__ValueKeyArray = [];
 		if (_array != null) this.addArray(_array);
 	}
 
-	function _isValidKey( _key )
+	function __isValidKey( _key )
 	{
 		return typeof _key == "string" && _key != "" && _key.slice(0,1).tolower() != _key.slice(0,1);
 	}
 
 	function add( _key )
 	{
-		if (!this._isValidKey(_key))
+		if (!this.__isValidKey(_key))
 		{
 			::logError("MSU Enum Keys must be non-empty strings and start with a capitalized letter");
 			throw ::MSU.Exception.InvalidValue(_key);
 		}
-		this._Keys[_key] <- this._Keys.len();
-		this._ValueKeyArray.push(_key);
+		this.__Keys[_key] <- this.__Keys.len();
+		this.__ValueKeyArray.push(_key);
 	}
 
 	function addArray( _array )
@@ -33,17 +34,17 @@
 
 	function getKeyForValue( _value )
 	{
-		return this._ValueKeyArray[_value];
+		return this.__ValueKeyArray[_value];
 	}
 
 	function len()
 	{
-		return this._Keys.len();
+		return this.__Keys.len();
 	}
 
 	function _get( _key )
 	{
-		if (_key in this._Keys) return this._Keys[_key];
+		if (_key in this.__Keys) return this.__Keys[_key];
 		throw null;
 	}
 }
--- a/scripts/config/msu/classes/enum.nut
+++ b/scripts/config/msu/classes/enum.nut
@@ -1,0 +1,49 @@
+::MSU.Class.Enum <- class
+{
+	_Keys = null;
+	_ValueKeyArray = null;
+
+	constructor( _array = null )
+	{
+		this._Keys = {};
+		this._ValueKeyArray = [];
+		if (_array != null) this.addArray(_array);
+	}
+
+	function _isValidKey( _key )
+	{
+		return typeof _key == "string" && _key != "" && _key.slice(0,1).tolower() != _key.slice(0,1);
+	}
+
+	function add( _key )
+	{
+		if (!this._isValidKey(_key))
+		{
+			::logError("MSU Enum Keys must be non-empty strings and start with a capitalized letter");
+			throw ::MSU.Exception.InvalidValue(_key);
+		}
+		this._Keys[_key] <- this._Keys.len();
+		this._ValueKeyArray.push(_key);
+	}
+
+	function addArray( _array )
+	{
+		foreach (value in _array) this.add(value);
+	}
+
+	function getKeyForValue( _value )
+	{
+		return this._ValueKeyArray[_value];
+	}
+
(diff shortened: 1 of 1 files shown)


scripts/config/msu/classes/enum.nut
```diff
@@ -0,0 +1,49 @@
...
+
+	function len()
+	{
+		return this._Keys.len();
+	}
+
+	function _get( _key )
+	{
+		if (_key in this._Keys) return this._Keys[_key];
+		throw null;
```
TaroEld, line 47: Why not throw InvalidKey?

TaroEld (commented):
One query otherwise lgtm

scripts/config/msu/classes/enum.nut
Enduriel: Sorry, forgot to point this out to you when you missed this, but basically https://github.com/MSUTeam/MSU/commit/50528f25488a5d93c5f2c8eec1f6c971275877fa. It means that squirrel will try and first check if there are built-in functions that match `_key` before throwing an error. It's in the doc for _get (2.15.2), so idk how I missed this before

scripts/config/msu/classes/enum.nut
```diff
@@ -0,0 +1,49 @@
+::MSU.Class.Enum <- class
+{
+	_Keys = null;
```
LordMidas, line 3: Afaik, we don't prefix class members with `_` so far in MSU. Do we want to start this convention now? My personal opinion is that it is unnecessary as by default we keep all members private and only expose them via getter setter functions as desired.

scripts/config/msu/classes/enum.nut
```diff
@@ -0,0 +1,49 @@
...
+	_ValueKeyArray = null;
+
+	constructor( _array = null )
+	{
+		this._Keys = {};
+		this._ValueKeyArray = [];
+		if (_array != null) this.addArray(_array);
+	}
+
+	function _isValidKey( _key )
```
LordMidas, line 13: Private functions should be prefixed with `__` i.e. two underscores. This is to differentiate them from metamethods.

scripts/config/msu/classes/enum.nut
```diff
@@ -0,0 +1,49 @@
+::MSU.Class.Enum <- class
+{
+	_Keys = null;
+	_ValueKeyArray = null;
+
+	constructor( _array = null )
+	{
+		this._Keys = {};
+		this._ValueKeyArray = [];
```
LordMidas, line 9: Might be more performant to have this as a table too with integers as its keys. Because `push` will normally resize the array to double the size (aka more memory is reserved) - not sure if this is how squirrel works but some programming languages do.

scripts/config/msu/classes/enum.nut
```diff
@@ -0,0 +1,49 @@
+::MSU.Class.Enum <- class
```
LordMidas, line 1: Why do you not allow removing an item?

LordMidas (changes requested):

scripts/config/msu/classes/enum.nut
Enduriel: I think we should start the convention to do that yes, (for private members specifically) also in this case it means that there are no keys that are not allowed in the enum.

scripts/config/msu/classes/enum.nut
Enduriel: That's fair

scripts/config/msu/classes/enum.nut
Enduriel: That could cause significant issues depending on when it was done and I can't think of any reason to do this.

scripts/config/msu/classes/enum.nut
Enduriel: tables are ~3x more memory intensive than arrays in squirrel according to electricimp

scripts/config/msu/classes/enum.nut
TaroEld: Yeah I think removing an item from an enum is a big nono

LordMidas: @Both:  I do see an issue that if someone removes an entry from the Enum and some other mod tries to access it, then yeah it will blow up. That'll blow up even in the case of the table stand-ins. So yeah, I agree, don't allow removing entries.  I guess I was talking more from a usability scope point of view - otherwise even adding items to an enum is bad (from a code perspective - as enums are primarily meant to be compile time).

Regarding usability, I consider this to be a replacement for the vanilla BB method of using Tables as stand-ns for enums. There the following things are important:
1. Adding/removing items from that table.
2. Cloning the table.
3. Iterating over the table.

Therefore, in my opinion, If this enum class is to offer a better alternative to that Table stand-in method, then it should be able to do these things as well (except perhaps the removal thing).

scripts/config/msu/classes/enum.nut
LordMidas: So far, we've had no public member in any class and I don't anticipate it happening any time soon. We use only getter/setter functions. This is the recommended way anyway. Public members are boilerplate and should be avoided. Public Properties in C# are getters/setters too. Therefore, I see this as redundant. All members should be private. Expose the ones you want to be public via getters/setters.

For functions, I agree that private ones should have a prefix.

scripts/config/msu/classes/enum.nut
LordMidas: Wow, that is interesting information. I suppose this should be kept in mind - because in certain places we are using tables to enhance "performance" but as BB is primarily a memory limited game, we should be careful about this. @TaroEld

scripts/config/msu/classes/enum.nut
LordMidas: So we agreed to add cloning and iteration?

scripts/config/msu/classes/enum.nut
Enduriel: @TaroEld disagrees with what you're saying here, therefore I'd rather be as explicit as possible

scripts/config/msu/classes/enum.nut
Enduriel: Yes, from what I remember we did wanna add cloning and iteration, I will take a look at doing that.

scripts/config/msu/classes/enum.nut
LordMidas: I understand that not having the prefix doesn't make it immediately obvious if this member is publicly accessible or not (via separately defined getter/setter functions) unless you check for the getter/setter functions' existence. However, the issue with your approach is that if tomorrow you want to expose any of these private members via a getter/setter function, you'd end up having to refactor a bunch of lines to remove the `__` prefix from the variable name.

scripts/config/msu/classes/enum.nut
```diff
@@ -0,0 +1,62 @@
...
+{
+	__Keys = null;
+	__ValueKeyArray = null;
+	__NextICache = null;
```
LordMidas, line 5: Unused variable

scripts/config/msu/string.nut
```diff
@@ -26,4 +26,14 @@
 		}
 		return true;
 	}
+
+	function startsWith( _string, _start )
```
LordMidas, line 30: This seems like a mistake i.e. changes to the strings utils file in this PR. Please remove the changes to this file.

LordMidas (changes requested):

scripts/config/msu/classes/enum.nut
Enduriel: Yes, there's not other way to do it since don't have a private function marker in squirrel. I don't see why this is an issue I think it's clean and standardized like this.

scripts/config/msu/string.nut
Enduriel: This was caused by the automatic base change, fixed by rebasing

LordMidas (approved):

TaroEld (approved):

PR #193, opened 2022-08-19
Settings tweaks

LordMidas:


Enduriel (approved):
lgtm

msu/squirrel_hooks/mod_hooks.nut
```diff
@@ -4,6 +4,7 @@
 	function add( _string )
 	{
 		::MSU.requireString(_string);
+		::logError(_string);
```
TaroEld, line 7: Is this intended?

msu/systems/mod_settings/abstract_setting.nut
```diff
@@ -87,7 +87,11 @@
 			{
 				::MSU.System.ModSettings.updateSettingInJS(this.getPanelID(), this.getID(), _value);
 			}
+
+			return true;
 		}
+
+		return _value == this.Value;
```
TaroEld, line 94: What's the reason to not return false here?

msu/systems/mod_settings/mod_settings_mod_addon.nut
```diff
@@ -39,4 +39,20 @@
...
 	}
+
+	function requireSettingValue( _setting, _value )
+	{
+		if (_setting.getValue() != _value)
+		{
+			if (_setting.isLocked())
+			{
+				::MSU.QueueErrors.add("Mod " + this.getModID() + " requires setting \'" + _setting.getID() + "\' of mod \'" + _setting.getModID() + "\' to have the value \'" + _value + "\' but it is locked to be \'" + _setting.getValue() + "\'. Lock reason: " + _setting.getLockReason());
+				return;
```
TaroEld, line 50: return false perhaps

msu/systems/mod_settings/mod_settings_mod_addon.nut
```diff
@@ -39,4 +39,20 @@
...
+			}
+
+			_setting.set(_value);
+		}
+
+		_setting.lock("Required by Mod: " + this.getModID());
```
TaroEld, line 56: And then return true here

msu/squirrel_hooks/mod_hooks.nut
Enduriel: Yes, if this doesn't print here it might not print at all if there are script errors in the code.

msu/systems/mod_settings/abstract_setting.nut
Enduriel: That's actually a good point, by the time it gets here this will always be false.

LordMidas: Because I want to return true if the already present value is equal to the passed _value.

msu/systems/mod_settings/mod_settings_mod_addon.nut
```diff
@@ -39,4 +39,29 @@
...
+		{
+			if (_setting.isLocked())
+			{
+				::MSU.QueueErrors.add("Mod " + this.getMod().getID() + " (" + this.getMod().getName() ") requires setting \'" + _setting.getID() + "\' of mod \'" + _setting.getMod().getID() + " (" + _setting.getMod().getName()")\' to have the value \'" + _value + "\' but it is locked to be \'" + _setting.getValue() + "\'. Lock reason: " + _setting.getLockReason() + ".");
+				return false;
+			}
+
+			if (_setting.set(_value))
+			{
+				_setting.lock("Required by Mod: " + this.getModID());
```
Enduriel, line 55: still using getModID (which doesn't exist)

Enduriel (changes requested):
=

msu/systems/mod_settings/mod_settings_mod_addon.nut
LordMidas: Yeah I fixed it while you were reviewing :p

msu/systems/mod_settings/abstract_setting.nut
```diff
@@ -122,13 +124,13 @@
...
 	{
-		return this.LockReason;
+		return this.LockReason == "" ? "" : this.LockReason.slice(0, -4);
 	}
 
 	function lock( _lockReason = "" )
 	{
 		this.Locked = true;
-		if (_lockReason != "") this.LockReason = _lockReason;
+		if (_lockReason != "") this.LockReason += _lockReason + "+++ ";
```
Enduriel, line 133: I personally think this will look awful, have you tested this?

msu/systems/mod_settings/abstract_setting.nut
```diff
@@ -122,13 +124,13 @@
...
 	{
-		return this.LockReason;
+		return this.LockReason == "" ? "" : this.LockReason.slice(0, -5);
 	}
 
 	function lock( _lockReason = "" )
 	{
 		this.Locked = true;
-		if (_lockReason != "") this.LockReason = _lockReason;
+		if (_lockReason != "") this.LockReason += _lockReason + " +++ ";
```
Enduriel, line 133: I still think this is gonna look really bad personally, could you share some screenshots?

Enduriel (changes requested):
Rest looks good though

msu/systems/mod_settings/abstract_setting.nut
```diff
@@ -68,8 +68,8 @@
 	{
 		if (this.Locked)
 		{
-			::logError("Setting \'" + this.Name + "\'' is locked and its value can't be changed");
-			return;
+			::logError("Setting \'" + this.Name + "\'' is locked and its value cannot be changed. Lock reason: " + this.getLockReason());
+			return false;
```
TaroEld, line 72: lockReason should probably also print the ID of the mod that locked it.

msu/systems/mod_settings/abstract_setting.nut
TaroEld, line 133: Should the reason be optional?

msu/systems/mod_settings/mod_settings_mod_addon.nut
```diff
@@ -39,4 +39,29 @@
...
 	}
+
+	function requireSettingValue( _setting, _value )
+	{
+		if (_setting.getValue() != _value)
+		{
+			if (_setting.isLocked())
+			{
+				::MSU.QueueErrors.add("Mod " + this.getMod().getID() + " (" + this.getMod().getName() + ") requires setting \'" + _setting.getID() + "\' of mod \'" + _setting.getMod().getID() + " (" + _setting.getMod().getName() + ")\' to have the value \'" + _value + "\' but it is locked to be \'" + _setting.getValue() + "\'. Lock reason: " + _setting.getLockReason() + ".");
+				return false;
```
TaroEld, line 50: Please use string formatting this is hideous. You can use getValue().tostring() if that is the issue.

LordMidas: Pretty sure `getValue().tostring()` throws an exception if `getValue()` returns a C++ instance e.g. Tile. We've run into this before in our `MSU.Exceptions` situation (see commit 4cd9b6ce520ad637850399b1e351cecfd8650151). So, I'm generally averse to using `format` anywhere where there might be anything whose value type is not certain.

msu/systems/mod_settings/abstract_setting.nut
LordMidas: On second thought, I think yeah the reason shouldn't be optional. But this is not changeable anymore as this is a released feature. @Enduriel What are your thoughts? Is there the possibility to deprecate and create a new function with reason not being optional?

LordMidas: Actually one way to achieve this would be to add a `lockSetting` function to `ModSettingsModAddon` which modders must use to lock other mod's settings. That way we can pass the locker's id which also fixes the issue that the locker's mod id is always printed in lock reason.

Currently people lock other mod's settings by directly accessing that setting and locking it. This isn't good.

Enduriel: > Currently people lock other mod's settings by directly accessing that setting and locking it. This isn't good.

I'm not sure I agree with this, I think that's fine.

I think this is overcomplicating the issue, if someone is locking a setting, they presumably know what they're doing, I don't think we should be trying to block/handhold to this extent.

TaroEld (approved):
Approved depending on last test of functionality

LordMidas:
I've tested it and it works. @Enduriel waiting for your review.

Enduriel (approved):

Issue #152, opened 2022-06-02
[BUG] ItemTypes added via the addNewItemType function get sorted at the top of the inventory window


LordMidas:
**Versions**
 - Game: [e.g. 1.5.0.11]
 - MSU: [e.g. 1.0.5]

**Describe the bug**
This is not a problem with our function, rather it also happens when you add new ItemTypes directly at the bottom of the ::Const.ItemType table. deDarxo was able to fix the sorting by adding an ItemType towards the top end of the table rather than at the bottom. Perhaps we should rework our functions so they add the stuff at the top end? Or we figure out what is causing the sorting problem and fix that.

Example:
This bug is causing Flails which are tagged as Cultist item to be sorted at the top of the inventory in Legends. That bug goes away if Cultist is added manually to the ItemType table towards the top end of the table (and push all values down by multiplying them with 2).

Enduriel:
I mean the issue here is that the vanilla sorting algorithm is weird.
```squirrel
if (_item1 == null && _item2 == null)
{
	return 0;
}
else if (_item1 == null && _item2 != null)
{
	return 1;
}
else if (_item1 != null && _item2 == null)
{
	return -1;
}
else if (_item1.getItemType() > _item2.getItemType())
{
	return -1;
}
else if (_item1.getItemType() < _item2.getItemType())
{
	return 1;
}
else if (_item1.getCategories() > _item2.getCategories())
{
	return -1;
}
else if (_item1.getCategories() < _item2.getCategories())
{
	return 1;
}
else if (_item1.getID() > _item2.getID())
{
	return -1;
}
else if (_item1.getID() < _item2.getID())
{
	return 1;
}
else
{
	return 0;
}
```
and it means that when we add a new ItemType it'll have the greatest value of all other ItemTypes, which means it'll affect sorting more than all the other itemtypes. Unfortunately this can only be solved by adding stuff to the bottom of the list, but even this still breaks sorting by Categories and IDs :/

PR #194, opened 2022-08-20
fix: itemTypes added via addNewItemType getting sorted to top (#152)

Enduriel:


msu/config/items.nut
```diff
@@ -38,15 +38,12 @@ foreach (itemType in ::Const.Items.ItemType)
 	}
 
 	local max = 0;
```
TaroEld, line 40: Leftover variable

TaroEld (approved):
Remove max then lgtm

msu/config/items.nut
```diff
@@ -37,16 +37,12 @@ foreach (itemType in ::Const.Items.ItemType)
 		return;
 	}
 
-	local max = 0;
-	foreach (w, value in ::Const.Items.ItemType)
-	{
-		if (value > max) max = value;
-	}
+	foreach (itemType, value in ::Const.Items.ItemType) ::Const.Items.ItemType[itemType] = value << 1;
```
LordMidas, line 40: Won't this cause `None = 0` to become `None = 1`?

Enduriel: No, because there are no 1 bits to shift into a 1 (it didn't happen during testing either)

LordMidas (approved):

PR #195, opened 2022-08-21
feat: add function to get colorpicker setting value as hex for sprite coloring

TaroEld:


TaroEld:
I know it's ugly

msu/systems/mod_settings/elements/color_picker_setting.nut
```diff
@@ -31,6 +31,19 @@
...
+	function getValueAsHexString()
+	{
+		local asArray = split(this.Value, ",");
+		local red = format("%x", asArray[0].tointeger());
+		if (asArray[0].tointeger() < 10) red = "0" + red;
+		local green = format("%x", asArray[1].tointeger());
+		if (asArray[1].tointeger() < 10) green = "0" + green;
+		local blue = format("%x", asArray[2].tointeger());
+		if (asArray[2].tointeger() < 10) blue = "0" + blue;
+		local opacity = asArray[3] == "0.0" ? "00" : format("%x", (asArray[3].tofloat() * 255).tointeger());
```
Enduriel, line 43: Doesn't this also need a < 10 check? right now this seems like it would still be wrong for opacity values that are between 1/255 and 9/255

Enduriel (changes requested):

msu/systems/mod_settings/elements/color_picker_setting.nut
TaroEld: indeed

Enduriel (approved):
lgtm, much cleaner now too

msu/systems/mod_settings/elements/color_picker_setting.nut
```diff
@@ -31,6 +31,20 @@
 		return asRGBA;
 	}
 
+	function getValueAsHexString()
+	{
+		local asArray = split(this.Value, ",");
+		local red = format("%x", asArray[0].tointeger());
+		local green = format("%x", asArray[1].tointeger());
+		local blue = format("%x", asArray[2].tointeger());
```
LordMidas, line 39: Have you tested if this works? It will throw an exception if `tointeger()` returns an integer instead of a string:
```
sq>print(format("%s", 2))

AN ERROR HAS OCCURED [string expected for the specified format]

CALLSTACK
*FUNCTION [main()] interactive console line [1]

LOCALS
[vargv] ARRAY
[this] TABLE
```

```
sq>print(format("%s", "2"))
2
sq>
```

```
sq>print(format("%s", "2".tointeger()))

AN ERROR HAS OCCURED [string expected for the specified format]

CALLSTACK
*FUNCTION [main()] interactive console line [1]

LOCALS
[vargv] ARRAY
[this] TABLE
```

LordMidas (changes requested):

msu/systems/mod_settings/elements/color_picker_setting.nut
TaroEld: %x, not %s. %x takes int and gives hex value.

LordMidas: Ah, I see. Then looks good.

LordMidas (approved):

PR #196, opened 2022-08-21
fix: vanilla bug where North and South roads can sometimes spawn but don't have brushes by giving them brushes

Enduriel:


msu/load.nut
```diff
@@ -29,3 +29,6 @@ includeLoad("hooks");
 includeLoad("msu_mod");
 includeLoad("vanilla_mod");
 includeLoad("test");
+
+this.Const.World.RoadBrushes.add(::Const.Direction.N, "road_N");
+this.Const.World.RoadBrushes.add(::Const.Direction.S, "road_S"); // idk where to put these
```
LordMidas, line 34: That is easy. Put them in `msu/config/world.nut` as that is the folder where we keep all our changes to vanilla `scripts/config` files.

msu/load.nut
```diff
@@ -29,3 +29,6 @@ includeLoad("hooks");
 includeLoad("msu_mod");
 includeLoad("vanilla_mod");
 includeLoad("test");
+
```
LordMidas, line 32: Please create a bug report for this on the vanilla forums and add a link to that in a comment here as we do for all other vanilla fixes.

LordMidas (changes requested):

msu/load.nut
Enduriel: https://steamcommunity.com/app/365360/discussions/1/3436829654723733817/

unpacked/world/metadata.xml
```diff
@@ -0,0 +1,4 @@
+<brush name="gfx/mods/msu/brushes/world.png" version="17">
```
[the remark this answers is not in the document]
Enduriel: I'd wanna know if your opinion on this has changed or if we need to discuss this properly, afaik @TaroEld likes unpacked as well.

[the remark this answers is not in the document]
TaroEld: Basically unpacked has worked before and works. That's pretty much the entirety of my opinion on it. <sub>Also I use it in my build scripts</sub>

Enduriel:
There are more missing brush that I'd wanna fix all at once (they're outlined in the vanilla bug report I made), this is minor and doesn't need to be done for 1.2

Enduriel:
This doesn't cover *all* the cases where this happens, a couple times I've seen some other strange shore brushes missing, but those are annoying to make (due to the existing art not really working for what we would need there and MUCH rarer compared to the ones I've fixed here.

unpacked_brushes/world/metadata.xml
LordMidas, line 1: I'd say change the brush name to `msu_world` just so that it doesn't conflict with any other file named `world.png`

LordMidas (approved):

unpacked_brushes/world/metadata.xml
Enduriel: We don't build to `/world.png` though, our brush goes to `/mods/msu/brushes/world.png`, which I don't think is likely to conflict lol

LordMidas: All brushes go directly in `gfx/` and in `brushes/` unless I am missing something.

Enduriel: You are, this isn't required, just vanilla convention

TaroEld (approved):

PR #198, opened 2022-08-22
fix(vanilla): skills not being added when a skill with the same ID which is garbage exists in SkillsToAdd

LordMidas:
I've tested the fix in PTR and it works. However, I think it will be more robust if we overwrite the vanilla `add` function and just add the `isGarbage` test as suggested in my vanilla bug report. This way we don't modify the skill ID and anything from a mod that may be checking for skill ids in `this.m.SkillsToAdd` won't get thrown off (although this is probably almost never going to happen).

EDIT: Check out the PR #206 as it has an implementation that accomplishes other things in addition to fixing this issue in different way.


TaroEld:
Closed in favor of #206

PR #200, opened 2022-09-07
feat: add item, weight foreach iteration

LordMidas:


Enduriel (approved):

Enduriel (changes requested):
Nvm, this doesn't actually work when tested with
```squirrel
local test = ::MSU.Class.WeightedContainer([
	[1, "A"],
	[5125, "B"]
]);
foreach (value in test) ::logInfo(value)
```

LordMidas:
> Nvm, this doesn't actually work when tested with
> 
> ```nut
> local test = ::MSU.Class.WeightedContainer([
> 	[1, "A"],
> 	[5125, "B"]
> ]);
> foreach (value in test) ::logInfo(value)
> ```

That is because the `::MSU.Table.keys` function needs to be fixed. It doesn't return anything atm. A fix is posted in another PR so you need code from both PRs to test it. I've tested the iteration on my own end and it works, but it'll be great if you can counter test.

Enduriel:
Yeah works for me after a rebase, you should do that to this PR though. You can't have an non-testable branch for a PR.
Besides that, I don't like that if you only iterate with 1 parameter, you're iterating over weights.

LordMidas:
Yeah I don't like item, weight either. I would've preferred weight, item. But I can't think of a way to achieve the latter. I think if we put this in the documentation then it is fine.

Enduriel:
Yeah... I guess there's nothing we can do about that. If you rebase to development, I'll approve the PR

Enduriel (approved):

TaroEld (approved):
lgtm assuming it works

PR #203, opened 2022-09-12
feat: add addBehavior utility function

LordMidas:
Allows writing stuff like:
```squirrel
addBehavior("BFAttackLunge", "Attack.BF_Lunge", 50, 100);
```
Instead of:
```squirrel
::Const.AI.Behavior.ID.BFAttackLunge <- id;
::Const.AI.Behavior.Name.push("Attack.BF_Lunge");
::Const.AI.Behavior.ID.COUNT = id + 1;
::Const.AI.Behavior.Order.BFAttackLunge <- 50;
::Const.AI.Behavior.Score.BFAttackLunge <- 100;
```

**Question:** Should we add it to `::Const.AI.` or should we create an `::MSU.AI.` utilities namespace?


TaroEld:
The latter probably

Enduriel (changes requested):
`::MSU.AI` is better imo

LordMidas:
I will move these to an `MSU.AI` namespace and merge it after #176 has been merged.

msu/config/ai.nut
```diff
@@ -0,0 +1,11 @@
+::Const.AI.addBehavior <- function( _id, _name, _order, _score )
+{
+	if (_id in ::Const.AI.Behavior.ID) throw ::MSU.Exception.DuplicateKey(_id);
+
+	::Const.AI.Behavior.ID[_id] <- ::Const.AI.Behavior.ID.COUNT;
+	::Const.AI.Behavior.ID.COUNT += 1;
+
+	::Const.AI.Behavior.Name.push(_name);
+	::Const.AI.Behavior.Order.RFAttackLunge <- _order;
+	::Const.AI.Behavior.Score.BFAttackLunge <- _score;
```
TaroEld, line 10: Should be 
```
::Const.AI.Behavior.Order[_id] <- _order;
::Const.AI.Behavior.Scorer[_id] <- _score;
```

TaroEld (changes requested):

PR #204, opened 2022-09-12
feat: add ability to specify a base item whose values are used for the named item before randomizing

LordMidas:
We should make a list of fields in `::MSU.Items.BaseWeaponFields` and iterate over that to allow easy way for modders to change the fields which are copied over from the Base Weapon. I will do this after #176 is merged (because of folder structure, to avoid resolving conflicts).

### DIFF
--- a/msu/hooks/items/armor/named/named_armor.nut
+++ b/msu/hooks/items/armor/named/named_armor.nut
@@ -1,0 +1,30 @@
+::MSU.HooksMod.hook("scripts/items/armor/named/named_armor", function(q) {
+	q.m.BaseItemScript <- null;
+
+	q.getValuesForRandomize <- function()
+	{
+		if (this.m.BaseItemScript == null)
+			return {};
+
+		local baseItem = ::new(this.m.BaseItemScript);
+		return {
+			Condition = baseItem.m.Condition,
+			ConditionMax = baseItem.m.ConditionMax,
+			StaminaModifier = baseItem.m.StaminaModifier
+		};
+	}
+
+	q.setValuesBeforeRandomize <- function( _values )
+	{
+		foreach (key, value in _values)
+		{
+			this.m[key] = value;
+		}
+	}
+
+	q.randomizeValues = @(__original) function()
+	{
+		this.setValuesBeforeRandomize(this.getValuesForRandomize());
+		return __original();
+	}
+});
--- a/msu/hooks/items/armor/named/named_armor.nut
+++ b/msu/hooks/items/armor/named/named_armor.nut
@@ -1,9 +1,12 @@
 ::MSU.HooksHelper.addBaseItemToNamedItem("scripts/items/armor/named/named_armor");
 
 ::MSU.HooksMod.hook("scripts/items/armor/named/named_armor", function(q) {
-	q.m.BaseItemFields = [
-		"Condition",
-		"ConditionMax",
-		"StaminaModifier"
-	];
+	q.getBaseItemFields = @() function()
+	{
+		return [
+			"Condition",
+			"ConditionMax",
+			"StaminaModifier",
+		];
+	}
 });
--- a/msu/hooks/items/armor/named/named_armor.nut
+++ b/msu/hooks/items/armor/named/named_armor.nut
@@ -1,4 +1,6 @@
 ::MSU.HooksMod.hook("scripts/items/armor/named/named_armor", function(q) {
+	::MSU.HooksHelper.addBaseItemToNamedItem(q);
+
 	q.getFieldsForRandomize <- function()
 	{
 		return [
@@ -9,6 +11,8 @@
 	}
 });
 
-::MSU.HooksMod.hookTree("scripts/items/armor/named/named_armor", function(q) {
-	::MSU.HooksHelper.addBaseItemToNamedItem(q);
+::MSU.QueueBucket.VeryLate.push(function() {
+	::MSU.HooksMod.hook("scripts/items/armor/named/named_armor", function(q) {
+		::MSU.HooksHelper.addBaseItemToNamedItemVeryLate(q);
+	});
 });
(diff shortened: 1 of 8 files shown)


TaroEld (changes requested):
Put this in a new function, like setBaseWeaponValues or something that is then called before randomizeValues() so that it can be amended/overwritten for weapons with special stats or something. Good idea

Enduriel (approved):

TaroEld:
postponed for 1.3 with potential changes

TaroEld (approved):
Approved, but for parity we should do the same for the other named item categories.

LordMidas:
I'm still not 100% happy with this. If a modder wants to add more fields that should be copied from the base item, the only way to do that right now would be to hook the `getValuesForRandomize` function, instantiate the base item, add the new values to the return of the hooked function, and then return that. 

We should instead try to have a function that returns an array of strings which are the fields that should be copied from the base item. Then modders can hook that function and modify the list - and the instantiation of the base item only happens once within the msu function that copies the values.

LordMidas:
> Okay, I understand the purpose now, 100% agree it's useful. I also think your last criticism is absolutely valid, though I'm less on board with your proposed solution. I think I'd prefer an array field or even global field for each class, something like that. I don't like this idea of infinitely hooking this function to add a single item to the array.

Yeah that was my original idea too but then @TaroEld proposed adding a function for that so that people have the possibility of modifying the stats for exotic weapons or if they want for niche cases - beyond just copying from the base item.

LordMidas:
Regarding using a global array for `getFieldsForRandomize` there is a slight "inconvenience" issue in that the fields are different for each named item type. So currently we'd need 3 separate global arrays for each of the following: named_weapon, named_shield, named_armor. If we just implement the function without a global array then each of these classes can have the list defined within that own custom definition of the function. So, with that in mind, what's the final word? Global arrays or keep inside function?

Enduriel:
please target #316 instead of development

Enduriel (commented):
Thoughts on creating a separate function to call inside of each of these which would apply the shared parts of the hooks? Specifically:
```
q.m.BaseItemScript <- null;
q.setValuesBeforeRandomize <- function( _values )
q.randomizeValues = @(__original) function()
```
?

Enduriel (approved):

Enduriel (approved):

msu/load.nut
```diff
@@ -7,7 +7,10 @@ local function includeLoad( _folder )
 
 includeLoad("ui");
 includeLoad("systems");
+includeLoad("hooks_helper");
```
TaroEld, line 10: Not sure why we have a folder and then a file that's named the same and include both ish

msu/hooks/items/weapons/named/named_weapon/named_weapon.nut
```diff
@@ -0,0 +1,22 @@
...
+			"RegularDamage",
+			"RegularDamageMax"
+			"ArmorDamageMult"
+			"ChanceToHitHead"
+			"DirectDamageMult"
+			"DirectDamageAdd"
+			"StaminaModifier"
+			"ShieldDamage"
+			"AdditionalAccuracy"
+			"FatigueOnSkillUse"
```
TaroEld, line 19: This would appear to be missing commas

msu/load.nut
TaroEld: But ok I guess you want to put it into different files later or sth.

TaroEld (changes requested):

msu/load.nut
Enduriel: yeah we might have more stuff I suspect, not a huge fan of how it's organized rn but hey that's for later us to fix when it becomes relevant I think.

TaroEld (approved):

msu/hooks/items/armor/named/named_armor.nut
```diff
@@ -0,0 +1,9 @@
+::MSU.HooksHelper.addBaseItemToNamedItem("scripts/items/armor/named/named_armor");
+
+::MSU.HooksMod.hook("scripts/items/armor/named/named_armor", function(q) {
+	q.m.BaseItemFields = [
+		"Condition",
+		"ConditionMax",
+		"StaminaModifier"
+	];
+});
```
TaroEld, line 9: Might as well pass the array into your helper and set it there.

TaroEld (changes requested):

TaroEld (approved):

Enduriel (approved):
This PR is pretty much a lesson in why composition >>> inheritance. Generally lgtm in this state.

Enduriel (approved):

TaroEld (approved):

msu/hooks_helper/hooks_helper.nut
```diff
@@ -0,0 +1,43 @@
...
+			q.setValuesBeforeRandomize <- function( _baseItem )
+			{
+				if (_baseItem != null)
+				{
+					foreach (field in this.getBaseItemFields())
+					{
+						this.m[field] = _baseItem.m[field];
+					}
+
+					this.m.ItemType = this.m.ItemType | ::Const.Items.ItemType.Named; // Re-add the named item type that got overwritten from the base item
```
Enduriel, line 21: Should be moved into a separate hookTree

Enduriel (changes requested):

msu/load.nut
```diff
@@ -7,6 +7,7 @@ local function includeLoad( _folder )
...
```
TaroEld, line 10: We discussed and we should probably expose this for when a mod adds another named item type?

LordMidas: Yes that sounds good to me.

LordMidas:
Postponed as I am going to first implement, test and polish this system over at the Modular Vanilla mod.

PR #205, opened 2022-09-13
feat: call onGetHitFactors on the target entity too to allow target's skills to modify the tooltip

LordMidas:
In order to identify within the onGetHitFactors function that this skill's actor is the target and not the user, one can check if `_skill.getContainer().getActor.getID() == this.getContainer().getActor().getID()` and if it is false then you know that this skill's actor is the target. Alternatively, check for `_targetTile.getEntity() != null && _targetTile.getEntity().getID() == this.getContainer().getActor().getID())`. 

The best solution would have been to add a new argument to the function, but unfortunately we cannot do that without breaking backwards compatibility. So I went with this route instead, which is fully backwards compatible.

Example of using the new functionality:
```squirrel
function onGetHitFactors( _skill, _targetTile, _tooltip )
{
	local user = _skill.getContainer().getActor();

	if (user.getID() == this.getContainer().getActor().getID())
	{
		_tooltip.push({
			icon = "ui/tooltips/positive.png",
			text = "Something positive!"
		});
	}
	else
	{
		_tooltip.push({
			icon = "ui/tooltips/negative.png",
			text = "Something negative!"
		});
	}
}
```


TaroEld (approved):

Enduriel (changes requested):
I'm worried about how this will affect skills that weren't built with this in mind? Also could this cause infinite recursion if pointed at your own tile? I think a different name for this function might be better.

LordMidas:
> I'm worried about how this will affect skills that weren't built with this in mind? Also could this cause infinite recursion if pointed at your own tile? I think a different name for this function might be better.

1. Good catch with the recursion. I've fixed that in a new commit.
2. No it won't affect any skills previously built. It is totally backward compatible.
3. I was hoping to avoid adding another function. One could add stuff like `onGetHitFactorsAsTarget` but I thought keeping everything inside one function is more readable as it keeps all the changes to HitFactors in one place.

LordMidas:
@Enduriel and @TaroEld So we can go about it in two different ways:
**Method 1**
Keep the current `onGetHitFactors` function and implement a new `onGetHitFactorsAsTarget` function which is called for the target entity. The advantage of this method is that the function for the skill user remains untouched and we add a new function for the target.

Usage:
```squirrel
// in my skill
function onGetHitFactors( _skill, _targetTile, _tooltip )
{
    _tooltip.push({
	icon = "ui/tooltips/positive.png", // positive chance to hit
	text = "Something positive!"
    });
}

function onGetHitFactorsAsTarget( _skill, _targetTile, _tooltip )
{
    _tooltip.push({
	icon = "ui/tooltips/negative.png", // negative chance to hit
	text = "Something negative!"
    });
}
```

**Method 2**
Mark the current `onGetHitFactors` function as deprecated and implement a new (name wip) function `onGetHitFactors2( _skill, _targetTile, _tooltip, _isTarget)` which is called for both user and target entity if target entity is different from user. Then inside this function the modder **MUST** use `if (!_isTarget)` to specify modifications to the tooltip from the user and `if (_isTarget)` to specify modifications from the target.

Usage:
```squirrel
function onGetHitFactors( _skill, _targetTile, _tooltip, _isTarget )
{
	if (!isTarget)
	{
		_tooltip.push({
			icon = "ui/tooltips/positive.png", // positive chance to hit
			text = "Something positive!"
		});
	}
	else
	{
		_tooltip.push({
			icon = "ui/tooltips/negative.png", // negative chance to hit
			text = "Something negative!"
		});
	}
}
```
What is preferable?

Enduriel:
I personally much prefer the 2 separate function approach, I don't really think there'll be enough skills affecting both your chance to hit something and that thing's chance to hit you for the confusion to be worth it personally. Additionally I really don't like appending an integer to a function name.

TaroEld:
Yeah, first approach for me too.

LordMidas:
I've pushed a fix. Will move all the MSU variables to a separate `.MSU.` table later in a separate refactor PR for all such files.

TaroEld (approved):

Enduriel (approved):

PR #206, opened 2022-09-15
feat: keep a history of how many times a non-stacking skill was added or removed and add or remove it accordingly

LordMidas:
This solves several problems.

## Advantage 1:
As highlighted in #198, during game deserialization if multiple skills unequip or equip an item in their `onAdded` function, the skills of the item are not added to the character. This implementation in this PR does not suffer from that issue. (Vanilla bug report: http://battlebrothersgame.com/forums/topic/weapon-skills-lost-on-load-if-multiple-skills-call-unequip-and-equip-in-onadded/)

## Advantage 2:
Problem: The game does not keep track of how many things add the same non-stacking skill to the container, and therefore, only the first addition and the last removal count. This leads to convoluted cases. See the steps below:
1. Let's say a character has two perks: Juggernaut and Indomitable. Juggernaut perk adds the Indomitable active skill while holding a two-handed weapon. Indomitable perk adds the Indomitable active skill unconditionally. 
2. The character picks the Juggernaut perk first, then equips a two-handed weapon, then picks the Indomitable perk.
3. This results in Indomitable not adding the Indomitable active skill because the character already has it and it is a non-stacking skill.
4. The character unequips the weapon, this causes Juggernaut to remove the Indomitable active skill.
5. The character has now lost the Indomitable active skill even though he still has the Indomitable perk.

The examples above are for proving the point only. This can extend to many other similar cases e.g. let's say a perk adds the Riposte skill unconditionally. Riposte is also added by Swords. A character equips a Sword, then picks that perk, then unequips the Sword, resulting in losing the Riposte skill. And so on.

The implementation in this PR solves this issue by keeping track of how many skills have "stacked" the addition of a non-stacking perk. When a skill is removed, it lowers this stack by 1 and the skill is only removed from the character when the stack reaches 0.

## Advantage 3:
This is going to be especially important with the feature in #163 that adds onEquip and onUnequip functions for skills which I have used extensively to add/remove skills based on the type of weapon equipped and the perk the event is being called on. I often run into the issue highlighted in Advantage 2 and am very interested in ironing this out.

## How it works:
Every skill now has a `MSU_AddedStack` variable which keeps track of how many times this skill has been added/removed. If the number of additions is greater than the number of removals i.e. the `MSU_AddedStack` is greater than 0, it sets the `.m.IsGarbage` variable of the skill to `false` so that it isn't removed during `collectGarbage`. 

A special case is handled for skills added by items. Items hold references to the skill objects they add. In this new method, if a skill is being added by an item to a character who already has a skill with the same id, we will only update the existing skill's `MSU_AddedStack`. Therefore, it is important to update the item's `SkillPtrs` array to now point to that existing skill instead of the skill object the item was adding and similarly update the `.m.Item` variable of that skill to point to that item.

### DIFF
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -56,6 +56,7 @@
 	o = o[o.SuperName];
 
 	o.m.MSU_AddedStack <- 1;
+	o.m.MSU_StackedFields <- {};
 
 	o.m.AIBehaviorID <- null;
 	o.m.DamageType <- ::MSU.Class.WeightedContainer();
@@ -337,9 +338,34 @@
 	local removeSelf = o.removeSelf;
 	o.removeSelf = function()
 	{
-		if (!this.isKeepingAddRemoveHistory()) return removeSelf();
+		this.removeSelfByStack(::MSU.Skills.StackedFields);
+	}
 
+	o.removeSelfByStack <- function( _stackedFields )
+	{
+		if (!this.isKeepingAddRemoveHistory()) return removeSelf();
 		if (--this.m.MSU_AddedStack == 0) return removeSelf();
+
+		foreach (fieldName, defaultValue in ::MSU.Skills.StackedFields)
+		{
+			local value = fieldName in _stackedFields ? _stackedFields[fieldName] : defaultValue;
+			local count = this.m.MSU_StackedFields[fieldName][value];
+
+			if (count > 0) this.m.MSU_StackedFields[fieldName][value] = count - 1;
+			else throw "trying to remove " + this.getID() + " but all its stacked additions with \'" + fieldName + " = " + value + "\' have already been removed";
+
+			if (this.m.MSU_StackedFields[fieldName][defaultValue] > 0) this.m[fieldName] = defaultValue;
+			else
+			{
+				local bestValue = defaultValue;
+				local bestCount = 0;
+				foreach (localValue, localCount in this.m.MSU_StackedFields[fieldName])
+				{
+					if (localCount > bestCount) bestValue = localValue;
+				}
+				this.m[fieldName] = bestValue;
+			}
+		}
 
 		// The actual item which provided this skill isn't unequipped yet because
 		// the removeSelf is called BEFORE the item is unequipped. So, we iterate over
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -364,9 +364,9 @@
 		this.setItem(null);
 	}
 
-	function isKeepingAddRemoveHistory()
+	o.isKeepingAddRemoveHistory <- function()
 	{
-		return !this.isStacking() && !(this.isType(::Const.SkillType.Perk) || !this.isType(::Const.SkillType.StatusEffect));
+		return !this.isStacking() && (this.isType(::Const.SkillType.Perk) || !this.isType(::Const.SkillType.StatusEffect));
 	}
 
 	o.getDamageType <- function()
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -339,12 +339,12 @@
 	{
 		if (!_skill.isKeepingAddRemoveHistory()) return removeSelf();
 
-		if (--this.m.MSU.AddedStack == 0) return removeSelf();
+		if (--this.m.MSU_AddedStack == 0) return removeSelf();
 
 		// The actual item which provided this skill isn't unequipped yet because
 		// the removeSelf is called BEFORE the item is unequipped. So, we iterate over
 		// all items and skip the one that is going to be unequipped
-		if (::MSU.isEqual(this.getContainer().getActor().getItems().m.MSU.ItemBeingUnequipped, this.getItem()))
+		if (::MSU.isEqual(this.getContainer().getActor().getItems().m.MSU_ItemBeingUnequipped, this.getItem()))
 		{
 			foreach (item in this.getContainer().getActor().getItems().getAllItems())
 			{
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -344,7 +344,7 @@
 		else
 		{			
 			if (--this.m.MSU.AddedStack <= 0) removeSelf();
-			else if (::MSU.isNull(this.getItem()) || ::MSU.isNull(this.getItem().getContainer()) || !::MSU.isEqual(this.getItem().getContainer().getActor(), this.getContainer().getActor()))
+			else if (::MSU.isNull(this.getItem()) || ::MSU.isNull(this.getItem().getContainer()) || this.getItem().getContainer().getActor().getID() != this.getContainer().getActor().getID())
 			{
 				foreach (item in this.getContainer().getActor().getItems().getAllItems())
 				{
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -341,7 +341,7 @@
 
 		if (--this.m.MSU.AddedStack == 0) return removeSelf();
 
-		if (::MSU.isNull(this.getItem()) || ::MSU.isNull(this.getItem().getContainer()) || this.getItem().getContainer().getActor().getID() != this.getContainer().getActor().getID())
+		if (::MSU.isNull(this.getItem()) || !::MSU.isEqual(this.getItem().getContainer(), this.getContainer().getActor().getItems()))
 		{
 			foreach (item in this.getContainer().getActor().getItems().getAllItems())
 			{
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -341,10 +341,15 @@
 
 		if (--this.m.MSU.AddedStack == 0) return removeSelf();
 
-		if (::MSU.isNull(this.getItem()) || !::MSU.isEqual(this.getItem().getContainer(), this.getContainer().getActor().getItems()))
+		// The actual item which provided this skill isn't unequipped yet because
+		// the removeSelf is called BEFORE the item is unequipped. So, we iterate over
+		// all items and skip the one that is going to be unequipped
+		if (::MSU.isEqual(this.getContainer().getActor().getItems().m.MSU.ItemBeingUnequipped, this.getItem()))
 		{
 			foreach (item in this.getContainer().getActor().getItems().getAllItems())
 			{
+				if (::MSU.isEqual(item, this.getItem())) continue;
+
 				foreach (skill in item.m.SkillPtrs)
 				{
 					if (skill.getID() == this.getID())
@@ -354,9 +359,9 @@
 					}
 				}
 			}
-
-			this.setItem(null);
 		}
+
+		this.setItem(null);
 	}
 
 	o.getDamageType <- function()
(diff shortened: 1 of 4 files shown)


Enduriel:
I thought this had issues but after writing out a scenario to highlight the issue I realized this handles it fine. I like this approach, especially because I didn't like #198. I still need to review the code more properly to approve the PR but theoretically I'm a fan.

Enduriel:
We need to standardize these rando members we're putting everywhere, I'm a big fan of adding an MSU table anytime we wanna add a member variable. You guys know this, can we commit to doing that? The only time this is an issue is when interfacing with JS, and member variables don't matter for this.

Therefore I'd suggest you switch from `MSU_AddedStack` to `MSU.AddedStack`

msu/hooks/skills/skill.nut
```diff
@@ -310,6 +311,38 @@
...
+	o.removeSelf = function()
+	{
+		if (this.isStacking())
+		{
+			removeSelf();
+		}
+		else
+		{
+			this.m.MSU_AddedStack--;
+			if (this.m.MSU_AddedStack <= 0) removeSelf();
```
Enduriel, line 324: I've noticed you don't do this in other places too, and I'd wanna review the codebase a bit for it, but part of the charm of in/decrement operators is that they never have to be their own line. Therefore this can be replaced with

```squirrel
if (--this.m.MSU_AddedStack <= 0) removeSelf();
```

msu/hooks/skills/skill_container.nut
```diff
@@ -27,6 +27,74 @@
...
+							_skillToAdd.setItem(null);
+							break;
+						}
+					}
+				}
+
+				if (_alreadyPresentSkill.m.MSU_AddedStack > 0)
+				{
+					_alreadyPresentSkill.m.IsGarbage = false;
+				}
```
Enduriel, line 55: When does this come up?

msu/hooks/skills/skill_container.nut
```diff
@@ -27,6 +27,74 @@
...
+
+			foreach (i, alreadyPresentSkill in this.m.SkillsToAdd)
+			{
+				if (alreadyPresentSkill.getID() == _skill.getID())
+				{
+					incrementAddedStack(alreadyPresentSkill, _skill);
+					return;
+				}
+			}
+		}
```
Enduriel, line 76: I think we should prioritize keeping the skill and item linked as much as possible, and only switch to a different item if the one currently linked is removed, or if the skill wasn't linked to an item in the first place.

Right now we swap items every time a new item is added with this skill.

Idk, I think either approach could cause issues but I feel like sticking with the same item is safer.

msu/hooks/skills/skill_container.nut
```diff
@@ -27,6 +27,74 @@
...
+		else _skill.removeSelf();
+	}
+
+	local removeByID = o.removeByID;
+	o.removeByID = function( _skillID )
+	{
+		local skill = this.getSkillByID(_skillID);
+		if (skill == null) return;
+
+		if (skill.m.MSU_AddedStack == 1) this.removeByID(_skillID);
```
Enduriel, line 94: Bad recursion on accident, remove `this.`

Enduriel (changes requested):
I forgot to start a review so sorry for the messy formatting

msu/hooks/skills/skill_container.nut
LordMidas: I've encountered this during `skill_container` deserialization whereby if a skill is added whose `onAdded` function unequips and equips the mainhand item. If **only one** such skill is present then there is no problem, but if multiple such skills are present then the skills of the weapon will be set to `Garbage` when the weapon is removed a second time. This happens even while the skills reside within `SkillsToAdd` array. See my bug report on vanilla forum from PR #198.

msu/hooks/skills/skill_container.nut
LordMidas: That approach is what I went with initially but it makes the entire thing **extremely** messy. Because items hold references directly to the skill objects, not ids. So we'd have to modify the `this.m.SkillPtrs` array of the new item and find the skill, remove it, and add the old skill. This also causes issues if the old skill was referenced elsewhere in the code. In this second iteration, it is much cleaner.

msu/hooks/skills/skill_container.nut
LordMidas: Nice catch

msu/hooks/skills/skill.nut
LordMidas: Done

msu/hooks/skills/skill_container.nut
Enduriel: Why would that be necessary? Can we not just not swap to a new item during onAdded if we already have one attached to the skill? And then during removeSelf swap to a new item if ours isn't equipped?

msu/hooks/skills/skill_container.nut
```diff
@@ -27,6 +27,74 @@
...
+			{
+				if (alreadyPresentSkill.getID() == _skill.getID())
+				{
+					incrementAddedStack(alreadyPresentSkill, _skill);
+					if (alreadyPresentSkill.m.MSU.AddedStack > 1) alreadyPresentSkill.onRefresh();
+					return;
+				}
+			}
+
+			foreach (i, alreadyPresentSkill in this.m.SkillsToAdd)
```
TaroEld, line 68: I don't know how these work, but by the name they make me think that they'll be `add()`-ed at their own time, so then this would increment twice maybe?

msu/hooks/skills/skill.nut
```diff
@@ -310,6 +314,35 @@
 		return ret;
 	}
 
+	local removeSelf = o.removeSelf;
+	o.removeSelf = function()
+	{
+		if (this.isStacking())
+		{
+			removeSelf();
```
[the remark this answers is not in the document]
TaroEld: idem for many following cases

TaroEld (changes requested):

msu/hooks/skills/skill_container.nut
LordMidas: Nah, they just get removed from the `this.m.SkillsToAdd` array and pushed to `this.m.Skills` array during the `collectGarbage` function. No new call to `add` function.

msu/hooks/skills/skill_container.nut
```diff
@@ -27,6 +27,74 @@
...
+			{
+				_alreadyPresentSkill.m.MSU.AddedStack += 1;
+				if (!::MSU.isNull(_skillToAdd.getItem()))
+				{
+					if (::MSU.isNull(_alreadyPresentSkill.getItem()) _alreadyPresentSkill.setItem(_skillToAdd.getItem());
+					foreach (i, itemSkill in _skillToAdd.getItem().m.SkillPtrs)
+					{
+						if (itemSkill.getID() == _skillToAdd.getID())
+						{
+							_skillToAdd.getItem().m.SkillPtrs[i] = _alreadyPresentSkill;
```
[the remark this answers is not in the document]
LordMidas: I thought about that but vanilla doesn't do weakrefs for this case, so I just followed vanilla convention.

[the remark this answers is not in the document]
Enduriel: There's no way for a skill to remain attached to an item without both of them being attached to an entity in vanilla, which is why I think this is a non-issue

msu/hooks/skills/skill_container.nut
```diff
@@ -27,6 +27,67 @@
...
+
+		foreach (alreadyPresentSkill in this.m.SkillsToAdd)
+		{
+			if (alreadyPresentSkill.getID() == _skill.getID())
+			{
+				incrementAddedStack(alreadyPresentSkill, _skill);
+				return;
+			}
+		}
+	}
```
Enduriel, line 72: There's a missing right bracket on line 41, but honestly after some more thought I really dislike this function to the point of thinking we should redo it entirely anyway to get rid of that closure.
There's also another mistake I think in that the original `add` isn't being run if this is the first time a stacking skill is being added (there should be a `return add` at the end)
Suggestion for rewrite, untested:
```squirrel
local add = o.add;
o.add = function( _skill, _order = 0 )
{
	if (_skill.isStacking()) return add(_skill, _order); 
	local i = 0;
	local allSkills = clone this.m.Skills
	allSkills.extend(this.m.SkillsToAdd)
	foreach (alreadyPresentSkill in allSkills)
	{
		++i;
		if (alreadyPresentSkill.getID() == _skill.getID())
		{
			if (++alreadyPresentSkill.m.MSU.AddedStack > 0) alreadyPresentSkill.m.IsGarbage = false;
			if (!::MSU.isNull(_skill.getItem()))
			{
				if (::MSU.isNull(alreadyPresentSkill.getItem())) alreadyPresentSkill.setItem(_skill.getItem());
				foreach (i, itemSkill in _skill.getItem().m.SkillPtrs)
				{
					if (itemSkill.getID() == _skill.getID())
					{
						_skill.getItem().m.SkillPtrs[i] = alreadyPresentSkill;
						_skill.setItem(null);
						break;
					}
				}
			}
			if (alreadyPresentSkill.m.MSU.AddedStack > 1 && this.m.Skills.len() <= i) alreadyPresentSkill.onRefresh();
			return;
		}
	}
	return add(_skill, _order);
}
```
I also just generally dislike the fact that there was a syntax error that would've come up during compilation in this PR and I'm gonna ask for some testing to be done to make sure all this works.

Enduriel (changes requested):

msu/hooks/skills/skill.nut
```diff
@@ -310,6 +314,31 @@
...
 	}
 
+	local removeSelf = o.removeSelf;
+	o.removeSelf = function()
+	{
+		if (this.isStacking()) return removeSelf();
+
+		if (--this.m.MSU.AddedStack <= 0) return removeSelf();
+
+		if (::MSU.isNull(this.getItem()) || ::MSU.isNull(this.getItem().getContainer()) || this.getItem().getContainer().getActor().getID() != this.getContainer().getActor().getID())
```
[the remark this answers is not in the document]
LordMidas: The `::MSU.isEqual` you talked about was about an item comparison because there can be multiple items on the same actor which have the same id (items in bags for example). Now we are doing an actor comparison and I don't see the need for an `isEqual` comparison as all actors have unique ids. Vanilla does this everywhere in the code for uniquely identifying actors - therefore it is in my opinion as reliable as it should be (otherwise what about all that vanilla code's reliabilty?) and more "familiar" and therefore less complicated.

msu/hooks/skills/skill.nut
```diff
@@ -310,6 +314,31 @@
 		return ret;
 	}
 
+	local removeSelf = o.removeSelf;
+	o.removeSelf = function()
+	{
+		if (this.isStacking()) return removeSelf();
+
+		if (--this.m.MSU.AddedStack <= 0) return removeSelf();
```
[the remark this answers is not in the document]
LordMidas: Fair

msu/hooks/skills/skill_container.nut
LordMidas: Yeah sorry the code got some mistakes after I refactored it based on Taro's review. Your rewrite looks like a good approach. I'll fix the thing.

LordMidas:
Closing this for now as we are shipping/testing this separately in https://github.com/Battle-Modders/stack-based-skills. Once a reliable implementation for this is established over there, then we can reconsider its implementation in MSU.

PR #207, opened 2022-09-20
fix: add check that the values in both tables are tables instead of only in one table

LordMidas:
This fixes the issue where it was only checking for one value being table instead of both. This `_recursively` thing was added by @Enduriel and personally I still don't know if it should be included at all. The function is supposed to be an analogy to `array.extend` and that function doesn't do recursive extension either. What do you guys think?

### DIFF
--- a/scripts/config/msu/table.nut
+++ b/scripts/config/msu/table.nut
@@ -39,7 +39,7 @@
 			if (key in _table1)
 			{
 				if (!_overwrite) throw ::MSU.Exception.DuplicateKey(key);
-				if (typeof value == "table" && _recursively)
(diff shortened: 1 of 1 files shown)


TaroEld (approved):

TaroEld:
> This fixes the issue where it was only checking for one value being table instead of both. This `_recursively` thing was added by @Enduriel and personally I still don't know if it should be included at all. The function is supposed to be an analogy to `array.extend` and that function doesn't do recursive extension either. What do you guys think?

I suppose recurse merge can be quite useful, though you could do it manually too.

Enduriel:
Yeah it's quite useful sometimes, and having a standardized way of doing it is nice.

Enduriel (approved):

PR #208, opened 2022-09-22
feat: add onSkillsUpdated event for skills

LordMidas:
This adds an event that is called at the end of `update()` after the `CurrentProperties` of the actor have been completely set up by the `update` function. It allows us to do things like the following:

```squirrel
// In some skill that is supposed to expire if the actor gets stunned
function onSkillsUpdated()
{
    if (this.getContainer().getActor().getCurrentProperties().IsStunned) this.removeSelf();
}
```
The above is currently impossible to do without hard coding like vanilla. Vanilla does it by using the `onAdded` function of the `Stunned` effect and any other effects e.g. `Horrified` that cause the character to be stunned and in that function they manually enter the ids of the skills that should be removed e.g. `effects.riposte`, `effects.spearwall` etc. This is very tightly coupled and very ugly for modders to try to mod - especially if a new status effect is added which also causes `IsStunned` in properties to be `true`.

The `MSU_callOnSkillsUpdated` variable is necessary because the `onSkillsUpdated` function of `actor` is called from within the skill container `update` function and because we call the `update` again at the end of the event, it will otherwise lead to infinite recursion.

Another thing it now enables is adding/removing skills during `onUpdate` and `onAfterUpdate` functions - as it triggers an update after the update.

The drawback of this is that it basically doubles the time of the `update` function (as it runs twice each time it is called now) - but I don't think that's practically a performance concern. An alternative, more efficient, implementation could be to instead have a `onAnySkillAdded( _skill )` function that is run after the `update` is run after a skill is added. However, this requires completely overwriting the vanilla `add` and `collectGarbage` functions of skill_container.

### DIFF
--- a/msu/hooks/entity/tactical/actor.nut
+++ b/msu/hooks/entity/tactical/actor.nut
@@ -1,5 +1,5 @@
 ::mods_hookExactClass("entity/tactical/actor", function(o) {
-	o.m.MSU_callOnSkillsUpdated <- false;
+	o.m.MSU_isCallingOnSkillsUpdated <- true;
 
 	local onMovementStart = o.onMovementStart;
 	o.onMovementStart = function ( _tile, _numTiles )
@@ -38,8 +38,8 @@
 		onSkillsUpdated();
 		if (!this.m.IsDying)
 		{
-			this.m.MSU_callOnSkillsUpdated = !this.m.MSU_callOnSkillsUpdated;
-			if (this.m.MSU_callOnSkillsUpdated) this.getSkills().onSkillsUpdated();
+			this.m.MSU_isCallingOnSkillsUpdated = !this.m.MSU_isCallingOnSkillsUpdated;
+			if (!this.m.MSU_isCallingOnSkillsUpdated) this.getSkills().onSkillsUpdated();
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


msu/hooks/entity/tactical/actor.nut
```diff
@@ -30,6 +32,17 @@
...
 	}
 
+	local onSkillsUpdated = o.onSkillsUpdated;
+	o.onSkillsUpdated = function()
+	{
+		onSkillsUpdated();
+		if (!this.m.IsDying)
+		{
+			this.m.MSU_callOnSkillsUpdated = !this.m.MSU_callOnSkillsUpdated;
+			if (this.m.MSU_callOnSkillsUpdated) this.getSkills().onSkillsUpdated();
```
TaroEld, line 42: Intuitively, the name is backwards, we don't really name variables in a way that tells us what to do.

Enduriel:
I'd like to have this variable be inside of an MSU table rather than MSU_VariableName, I believe that's much cleaner

msu/hooks/entity/tactical/actor.nut
LordMidas: Check now and lemme know if you are ok with the change.

Enduriel (approved):

LordMidas:
I've pushed a fix. Will move all the MSU variables to a separate .MSU. table later in a separate refactor PR for all such files.

TaroEld (approved):

Enduriel (approved):
Assuming the msu table stuff does get changed later

PR #212, opened 2022-10-21
fix: change weapontype of vanilla slings from throwing to sling

### LINKED ISSUE #210: [REQUEST] Correct the Weapontype of Nomad Sling and Staff Sling
In Vanilla the weapons **Nomad Sling** and **Staff Sling** are categorized as **Throwing Weapon**.
VanillaSlowingWeaponCat

So if I use MSU then it will translate those categories into making those weapons have the **Throwing Weapon** weapontype.
This can be seen in this screenshot:
SlingCategorizedAsThrowingWeapon

Why ship MSU with **Sling** as a default weapontype when it is never used in Vanilla?
After all you allow us to add new weapontypes outselves.
WhyHaveSLingInTheFirstPlace

Refactor the weapontypes of **Nomad Sling** and **Staff Sling** to become **Slings** instead of **Throwing Weapons**

**Describe alternatives you've considered**
The only alternative to make this consistent would be removing **Sling** from these weapontypes. But this would break mods that previously expected this weapontype to already exist.

Darxo:
Closes #210

### DIFF
--- a/msu/hooks/items/weapons/oriental/nomad_sling.nut
+++ b/msu/hooks/items/weapons/oriental/nomad_sling.nut
@@ -1,0 +1,7 @@
+::mods_hookNewObject("items/weapons/oriental/nomad_sling", function(o) {
+	if (o.isWeaponType(::Const.Items.WeaponType.Throwing))
+	{
+		o.removeWeaponType(::Const.Items.WeaponType.Throwing, false);
+		o.addWeaponType(::Const.Items.WeaponType.Sling);
+	}
+});
--- a/msu/hooks/items/weapons/staff_sling.nut
+++ b/msu/hooks/items/weapons/staff_sling.nut
@@ -1,0 +1,7 @@
+::mods_hookNewObject("items/weapons/staff_sling", function(o) {
+	if (o.isWeaponType(::Const.Items.WeaponType.Throwing))
+	{
+		o.removeWeaponType(::Const.Items.WeaponType.Throwing, false);
+		o.addWeaponType(::Const.Items.WeaponType.Sling);
+	}
+});


Enduriel:
This one is a little more complicated, but I think this is the only way to do this, otherwise skills written for throwing weapons would also apply to slings. Technically we are changing vanilla tooltips but eh from my pov.

Enduriel (approved):
Code looks good too, though now it made me wonder if it's more efficient to wrap a create function with hookExactClass or use a hookNewObject

LordMidas:
I disagree with this PR because MSU should mimic vanilla behavior.

LordMidas (changes requested):
I disagree with this PR because MSU should mimic vanilla behavior.

Enduriel:
> I disagree with this PR because MSU should mimic vanilla behavior.

This is also true, but how do you expect stuff that affects only Staffs to use MSU for this? what if we have a bunch of different mods adding a Staff WeaponType. I don't think either option here is perfect tbh. Staffs are badly handling in vanilla I think.

LordMidas:
> > I disagree with this PR because MSU should mimic vanilla behavior.
> 
> This is also true, but how do you expect stuff that affects only Staffs to use MSU for this? what if we have a bunch of different mods adding a Staff WeaponType. I don't think either option here is perfect tbh. Staffs are badly handling in vanilla I think.

My argument would be that that's the mod's job. MSU is not responsible for creating a weapon type for each possible weapon type that mods could use - we are only responsible for creating the weapontype framework. However, I think it is too late to remove these weapontypes from MSU now because this might break savegames for mods who depend on these. So, we keep them in for now, but I still believe we shouldn't change any weapon type of vanilla weapons in MSU.

Enduriel:
Okay, that seems reasonable to me. @TaroEld, thoughts?

TaroEld:
I guess, does it not use the current type for something?

TaroEld:
My previous comment was made without me seeing midas' comments somehow, and yeah his point sounds reasonable.

TaroEld:
We can talk about this again later

PR #213, opened 2022-10-21
fix: add check to prevent slice on empty string

### LINKED ISSUE #211: [BUG] setupCategories crashes the game while a weapon has no weaponTypes
**Versions**
 - MSU: 1.1.1

**Describe the bug**
When the setupCategories function of a weapon with no weapontypes is called then the game crashes.
This happens for example if I want to remove the last existing weapontype on a weapon because I want to replace it with another weapontype. But I forget to add the second parameter `false` on the removeWeaponType function

**To Reproduce**

1. Remove the weaponType of a weapon that has only one weapontype like in the screenshots below
2. See error

CrashRemoveLastWeaponType

**Expected behavior**
Removing the last weapontype of a weapon should never cause errors. The only reason this currently is the case because you try to slice some characters from the end of an empty string

**Log**
CrashRemoveLastWeaponTypeLog

Darxo:
I tested this fix ingame for a weapon.
Now you are able to have a Weapon with no weapontype (just one-handed or two-handed by themselves)
But more importantly: The game doesn't crash anymore if someone removes the last type of an item even if they give it another type right after.

Closes #211

### DIFF
--- a/msu/hooks/items/weapons/weapon.nut
+++ b/msu/hooks/items/weapons/weapon.nut
@@ -127,7 +127,7 @@
 			}
 		}
 
-		this.m.Categories = this.m.Categories.slice(0, -1) + ", ";
+		if (this.m.Categories != "") this.m.Categories = this.m.Categories.slice(0, -1) + ", ";
 
 		if (this.isItemType(::Const.Items.ItemType.OneHanded))
 		{


Enduriel:
@TaroEld @LordMidas lgtm I'm happy to just merge it in as is

Enduriel (approved):

msu/hooks/items/item.nut
```diff
@@ -60,10 +60,11 @@
 			}
 		}
 
-		return names.len() > 1 ? "[color=" + ::Const.UI.Color.NegativeValue + "]" + names.slice(0, -2) + "[/color]\n\n" + getDescription() : getDescription();
+		if (names != "") names = names.slice(0, -2);
```
LordMidas, line 63: This change is wrong. The original code is fine as far as I can see unless I am missing something?

LordMidas (changes requested):

msu/hooks/items/item.nut
Enduriel: the one thing I think we should change would be instead of `names.len() > 1` we should use `names != ""` since I you measured the performance to be 2x better (I just double checked this and yeah that is correct), but other than that yeah looking back I don't see what the problem is with this portion

Enduriel (changes requested):
Not sure that the changes to item.nut are actually necessary anymore

msu/hooks/items/item.nut
Darxo: Yeah this didn't need that check, my bad

Enduriel (approved):

LordMidas (approved):

PR #215, opened 2022-10-22
fix: range setting being unable to handle floats

Enduriel:
All of this works fine from my testing and is in theory good as is. The limitation here is that we can only handle floats with up to 6 significant figures, and this isn't communicated to the modder in any way as it I'd have to write some sort of significant figure rounder for squirrel floats (doesn't exist in base squirrel or the vanilla math library) to check if the float passed has less than 6 significant figures.

It would also have to test if adding/removing step from min/max/value could lead to a value with more than 6 significant figures.

This is all possible, but quite complicated. For now I think this addresses one of the main known bugs we've had for a while and we should try and push this with 1.2 no matter what, with potentially working on parsing the input and informing the modder if they are passing unacceptable inputs later.

In practice this is very unlikely, the UI can't even really handle values of 3 significant figures assuming the full range is used with the highest level of granularity.

### DIFF
--- a/msu/systems/mod_settings/mod_settings_system.nut
+++ b/msu/systems/mod_settings/mod_settings_system.nut
@@ -79,17 +79,24 @@
 		/*
 		_data = {
 			modID = {
-				settingID = value
+				settingID =
+				{
+					type,
+					value
+				}
 			}
 		}
 		*/
-
 		foreach (modID, panel in _data)
 		{
-			foreach (settingID, value in panel)
+			foreach (settingID, data in panel)
 			{
-				local setting = this.Panels[modID].getSetting(settingID);
-				setting.set(value);
+				this.updateSettingFromJS({
+					mod = modID,
+					id = settingID,
+					type = data.type,
+					value = data.value
+				});
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


msu/systems/mod_settings/mod_settings_system.nut
```diff
@@ -79,17 +79,24 @@
...
+			foreach (settingID, data in panel)
 			{
-				local setting = this.Panels[modID].getSetting(settingID);
-				setting.set(value);
+				this.updateSettingFromJS({
+					mod = modID,
+					id = settingID,
+					type = data.type,
+					value = data.value
+				});
```
TaroEld, line 99: I don't like how this calls the JS function, when it might as well be called from SQ, Probably just add an universial updateSetting function.

Enduriel: This function is called from JS though, maybe this function should be renamed to updateSettingsFromJS instead, I'd prefer that approach

TaroEld (approved):
We can change the name of that other function down the line I suppose, doesn't entirely fit this PR

LordMidas (approved):
I don't understand the JS side so looks fine to me.

PR #221, opened 2022-11-01
feat: add uncapitalizeFirst function to ::MSU.String

Enduriel:
This should be merged after update-checker is merged in, I've set it up this way because I'm using this in a current project and I need both this and update-checker

### DIFF
--- a/scripts/config/msu/string.nut
+++ b/scripts/config/msu/string.nut
@@ -10,9 +10,7 @@
 	function uncapitalizeFirst( _string )
 	{
 		if (_string == "") return _string;
-		local first = _string.slice(0, 1);
-		first = first.tolower();
-		return first + _string.slice(1);
+		return _string.slice(0, 1).tolower() + _string.slice(1);
 	}
 
 	function replace( _string, _find, _replace, _all = false )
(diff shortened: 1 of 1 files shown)


scripts/config/msu/string.nut
```diff
@@ -7,6 +7,14 @@
 		return first + _string.slice(1);
 	}
 
+	function uncapitalizeFirst( _string )
+	{
+		if (_string == "") return _string;
+		local first = _string.slice(0, 1);
```
LordMidas, line 13: `local first = _string.slice(0, 1).tolower();` instead of a second line.

LordMidas (approved):
Change to that one liner. No big deal if you don't.

scripts/config/msu/string.nut
```diff
@@ -7,6 +7,12 @@
 		return first + _string.slice(1);
 	}
 
+	function uncapitalizeFirst( _string )
+	{
+		if (_string == "") return _string;
```
LordMidas, line 12: One liners rule the world:
`return _string == "" ? _string : _string.slice(0, 1).tolower() + _string.slice(1);`

TaroEld (approved):

PR #223, opened 2022-11-03
fix: refactor the query tooltip functions to include a call to the base vanilla function instead of complete overwrite

LordMidas:
This fixes the issues caused by these functions having been updated in vanilla since the time the original hook in MSU was written leading to wrong behavior. This new method is less efficient as it calls getEntityById and getSkillByID while these things are also called by the vanilla functions, but this implementation is more robust against vanilla updates to these functions.

### DIFF
--- a/msu/hooks/ui/screens/tooltip/tooltip_events.nut
+++ b/msu/hooks/ui/screens/tooltip/tooltip_events.nut
@@ -13,41 +13,28 @@
 	local general_querySkillTooltipData = o.general_querySkillTooltipData;
 	o.general_querySkillTooltipData = function( _entityId, _skillId )
 	{
-		local entity = ::Tactical.getEntityByID(_entityId);
+		local ret = general_querySkillTooltipData(_entityId, _skillId);
 
-		if (entity != null)
+		if (ret != null)
 		{
-			local skill = entity.getSkills().getSkillByID(_skillId);
-
-			if (skill != null)
-			{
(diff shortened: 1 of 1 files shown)


Enduriel (approved):

msu/hooks/ui/screens/tooltip/tooltip_events.nut
```diff
@@ -13,41 +13,28 @@
...
-			local skill = entity.getSkills().getSkillByID(_skillId);
-
-			if (skill != null)
-			{
-				local tooltip = skill.getTooltip();
-				skill.getContainer().onQueryTooltip(skill, tooltip);
-				return tooltip;
-			}
+			local skill = ::Tactical.getEntityByID(_entityId).getSkills().getSkillByID(_skillId);
+			skill.getContainer().onQueryTooltip(skill, ret);
```
TaroEld, line 21: Probably just add the same checks as the base function (entity != null && skill != null). It's technically not necessary as the base function only returns non-null if those conditions aren't met, but there might be mods that redirect things in that case.

TaroEld: Actually it's probably not important, we should come first

LordMidas: I don't think anyone should change that function to return a non-null value if either entity or skill is null because that's the whole point of the function.

PR #224, opened 2022-11-05
feat: add support for individuel ParagonLevel

### LINKED ISSUE #218: [REQUEST] new MaxLevelWithPerkpoints variable for player.nut
The last level where you receive a perk point in vanilla is 11. This is controlled by the `::Const.XP.MaxLevelWithPerkpoints` variable.
But there is no way currently to have different values for different brothers. So that some have a smaller max level than others.
Different MaxLevelWithPerkpoints for different backgrounds are a useful tool for custom origins (like Manhunter) or for new game mechanics where through luck or by buying expensive backgrounds you sometimes have a higher `MaxLevelWithPerkpoints` on those brothers.

Introduce a new `m.MaxLevelWithPerkpoints` variable in the `player.nut`. 
This variable will mirror `::Const.XP.MaxLevelWithPerkpoints` during the `function create()` and then it can be overwritten by custom scenarios or difficulties.
Hook the `function updateLevel()` of the `player.nut` and switch out the value of the old `::Const.XP.MaxLevelWithPerkpoints` global variable with the value of the newly introduced `this.m.MaxLevelWithPerkpoints`. Afterwards switch those values back.

Darxo:
- new ParagonLevel variable for player.nut that defines when they stop gaining perks
- fix instances of hardcoded mention of the ParagonLevel in tooltip
- fix hardcoded ui coloring of ParagonLevel
- Closes #218

### DIFF
--- a/msu/hooks/entity/tactical/player.nut
+++ b/msu/hooks/entity/tactical/player.nut
@@ -1,10 +1,48 @@
 ::mods_hookExactClass("entity/tactical/player", function(o) {
 	o.m.LevelUpsSpent <- 0;
+	o.m.ParagonLevel <- 11;
+
+    local oldCreate = o.create;
+    o.create = function()
+    {
+        oldCreate();
+        this.m.ParagonLevel = ::Const.XP.MaxLevelWithPerkpoints;  // Global default value in vanilla
+    }
 
 	o.getMovementSpeedMult <- function()
 	{
 		return 1.0;
 	}
+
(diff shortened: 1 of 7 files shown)


Darxo:
This implementation still misses a good way to retroactively change the ParagonLevel of a brother (outside of the create() function).
Because changing it means having to flush & recalculate the AttributeLevelUpValues.

This PR is more for discussing and not 100% thought through and finished implementation.

PR #227, opened 2022-11-08
fix(vanilla): fix skills sometimes not being added properly

Enduriel:
The scenario where this could occur is outlined here http://battlebrothersgame.com/forums/topic/weapon-skills-lost-on-load-if-multiple-skills-call-unequip-and-equip-in-onadded.

This is a nicer alternative to #198.

### DIFF
--- a/msu/hooks/skills/skill_container.nut
+++ b/msu/hooks/skills/skill_container.nut
@@ -34,6 +34,25 @@
 		}
 	}
 
+	// VANILLAFIX http://battlebrothersgame.com/forums/topic/weapon-skills-lost-on-load-if-multiple-skills-call-unequip-and-equip-in-onadded
+	// if we overwrite add later, we should do this inside the loop in the function
+	local add = o.add;
+	o.add = function( _skill, _order = 0 )
+	{
+		if (!_skill.isStacking())
+		{
+			foreach (skill in this.m.SkillsToAdd)
+			{
+				if (_skill.getID() == skill.getID() && skill.isGarbage())
+				{
+					this.m.SkillsToAdd.remove(skill);
+					break;
+				}
+			}
+		}
+		return add(_skill, _order);
+	}
+
 	o.callSkillsFunction <- function( _function, _argsArray = null, _update = true, _aliveOnly = false )
 	{
 		if (_argsArray == null) _argsArray = [null];


msu/hooks/skills/skill_container.nut
```diff
@@ -34,6 +34,25 @@
...
 	}
 
+	// VANILLAFIX http://battlebrothersgame.com/forums/topic/weapon-skills-lost-on-load-if-multiple-skills-call-unequip-and-equip-in-onadded
+	// if we overwrite add later, we should do this inside the loop in the function
+	local add = o.add;
+	o.add = function( _skill, _order = 0 )
+	{
+		if (!_skill.isStacking())
+		{
+			foreach (skill in this.m.SkillsToAdd)
```
TaroEld, line 44: Call it `skillToAdd` instead of `skill` for readability

msu/hooks/skills/skill_container.nut
```diff
@@ -34,6 +34,25 @@
...
+			{
+				if (_skill.getID() == skill.getID())
+				{
+					if (skill.isGarbage()) this.m.SkillsToAdd.remove(skill);
+					break;
```
TaroEld, line 49: Why break, can there not be multiple to add?

TaroEld (changes requested):

msu/hooks/skills/skill_container.nut
LordMidas: `alreadyPresentSkill` is better as `skillToAdd` is confusing with `_skill` which is the skill to be added.

msu/hooks/skills/skill_container.nut
```diff
@@ -34,6 +34,25 @@
...
+	// VANILLAFIX http://battlebrothersgame.com/forums/topic/weapon-skills-lost-on-load-if-multiple-skills-call-unequip-and-equip-in-onadded
+	// if we overwrite add later, we should do this inside the loop in the function
+	local add = o.add;
+	o.add = function( _skill, _order = 0 )
+	{
+		if (!_skill.isStacking())
+		{
+			foreach (skill in this.m.SkillsToAdd)
+			{
+				if (_skill.getID() == skill.getID())
```
LordMidas, line 46: Just use a single if statement. The && operator only evaluates the second condition if the first condition is true therefore these two if statements can be combined into one: `if (_skill.getID() == skill.getID() && skill.isGarbage())`

msu/hooks/skills/skill_container.nut
```diff
@@ -34,6 +34,25 @@
...
+				{
+					if (skill.isGarbage()) this.m.SkillsToAdd.remove(skill);
```
LordMidas, line 48: Wrong use of `remove` - it expects an index. Therefore you need to iterate with an index and use `remove(i)`.

LordMidas (changes requested):

msu/hooks/skills/skill_container.nut
LordMidas: There won't be multiples because on every addition of that skill type it will check and remove it.

LordMidas:
Also please make the commit message more specific - something like: `fix(vanilla): skills not being added if a skill with the same id is garbage and exists in SkillsToAdd`

LordMidas:
Commit message for PR:
```
fix(vanilla): skills sometimes not being added properly

Specifically, if a skill with the same id is garbage and exists in SkillsToAdd.
```

LordMidas (approved):

TaroEld (approved):

PR #229, opened 2022-11-11
feat: deprecate MSU.Utils.addAIBehaviour

LordMidas:
This reverts commit b6e45896b716120fc69dbdddf3ef9bd496af692c. A function with this functionality was added to ::MSU.AI namespace instead in #203.

### DIFF
--- a/scripts/config/msu/utils.nut
+++ b/scripts/config/msu/utils.nut
@@ -218,14 +218,9 @@
 	    return this.Timers[_id];
 	}
 
+	// Deprecated - use ::MSU.AI.addBehavior instead
 	function addAIBehaviour(_id, _name, _order, _score = null)
 	{
-		::Const.AI.Behavior.ID[_id] <- ::Const.AI.Behavior.ID.COUNT++;
-		::Const.AI.Behavior.Name.push(_name);
-		::Const.AI.Behavior.Order[_id] <- _order;
-		if(_score != null)
-		{
-			::Const.AI.Behavior.Score[_id] <- _score;
-		}
+		return ::MSU.AI.addBehavior(_id, _name, _order, _score);
 	}
 }


Enduriel:
This was added for 1.1 and was documented in the changelog, we need to update the wiki but this makes me realize that #203 may be redundant

LordMidas:
> This was added for 1.1 and was documented in the changelog, we need to update the wiki but this makes me realize that #203 may be redundant

Oh I see. Yeah that is unfortunate. I still would rather keep this in a separate namespace i.e. MSU.AI because I don't like the idea of polluting the utils namespace with niche functions whose scope is limited to very niche cases or which belong in namespaces which we may expand upon in the future.

Enduriel:
Okay yeah that's fine by me, if you point this at the function added in #203 instead of removing it I'll approve.

LordMidas:
@Enduriel I've updated the PR. I've deprecated the old function and pointed it at the new one. However, the old one allows `_score` to be null, which is weird imo - the new function doesn't handle null `_score` as I don't think an AI behavior should ever have null score. @TaroEld Could you please explain why you allowed null `_score` in your implementation?

TaroEld:
> @Enduriel I've updated the PR. I've deprecated the old function and pointed it at the new one. However, the old one allows `_score` to be null, which is weird imo - the new function doesn't handle null `_score` as I don't think an AI behavior should ever have null score. @TaroEld Could you please explain why you allowed null `_score` in your implementation?

I'm fine with requiring score, but the game doesn't actually need it. There are 83 behaviours excluding None, and only 80 score excluding Zero, so the game also doesn't use them at some points.

Enduriel (approved):

TaroEld (approved):

PR #230, opened 2022-11-12
feat: replace getModID function with getMod

LordMidas:


TaroEld (approved):

Enduriel (changes requested):
No, if we're doing it this way we should be storing a reference to the mod, not the modID. The ID takes up more memory and is slower to turn into a mod. Do the conversion in the constructor.

Enduriel (approved):

PR #232, opened 2022-11-13
feat: add methods to pause and unpause to SQ and JS timer classes

TaroEld:
This can help time the execution of a specific component more accurately.

### DIFF
--- a/scripts/config/msu/classes/timer.nut
+++ b/scripts/config/msu/classes/timer.nut
@@ -12,20 +12,21 @@
 		this.PauseIncrement = 0;
 	}
 
-	function get( _msg = "", _stop = false, _printMsg = true )
+	function silentGet()
 	{
-		// Make sure to substract paused timme
 		local pauseIncrement = this.PauseIncrement;
 		if (this.PauseStart != null)
 			pauseIncrement += ::Time.getExactTime() - this.PauseStart;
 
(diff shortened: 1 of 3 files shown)


scripts/config/msu/classes/timer.nut
```diff
@@ -2,22 +2,49 @@
...
 		this.Start = ::Time.getExactTime();
+		this.PauseIncrement = 0;
 	}
 
 	function get(_msg = "", _stop = false)
 	{
-	    local time = (::Time.getExactTime() - this.Start) * 1000;
+		// Make sure to substract paused timme
+		if (this.PauseStart != null)
+			this.unpause();
```
TaroEld, line 19: This is a bit of a question, since technically you can get() the current time at any point, and now this would unpause the timer. Instead it could just substract the incremented time or something. Suggestions?

Enduriel: use a local variable here for pauseIncrement that you set differently based on whether the timer is currently paused

Enduriel (changes requested):
I think adding a function that calls `get("", false, _printMsg)` and one that calls `stop("", _printMsg)` would be a good call. Maybe something like silentGet and silentStop?

Enduriel (approved):

LordMidas (approved):

PR #235, opened 2022-11-14
Split Settings callback into callbacks for Before and After the value was changed

TaroEld:
Originally, the callbacks ran before the value was changed, passing the new value.  This was not very convenient in practice. With this PR, the user can decide whether to run the callback before or after a change. The new value gets passed in BeforeChange callbacks, the old value in AfterChange callbacks.

### DIFF
--- a/msu/systems/mod_settings/abstract_setting.nut
+++ b/msu/systems/mod_settings/abstract_setting.nut
@@ -5,7 +5,8 @@
 	BaseValue = null;
 	Locked = null;
 	LockReason = null;
-	Callbacks = null;
+	BeforeChangeCallbacks = null;
+	AfterChangeCallbacks = null;
 	Persistence = null; //if it should print change to log for further manipulation
 
 	constructor( _id, _value, _name = null, _description = null )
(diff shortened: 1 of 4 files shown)


msu/systems/mod_settings/abstract_setting.nut
```diff
@@ -64,28 +79,33 @@
 		this.set(_value, _updateJS, _updatePersistence, _updateCallback, _force);
 	}
 
-	function set( _value, _updateJS = true, _updatePersistence = true, _updateCallback = true, _force = false)
+	function set( _newValue, _updateJS = true, _updatePersistence = true, _updateBeforeChangeCallback = true, _force = false, _updateAfterChangeCallback = true)
```
LordMidas, line 82: Oh boy this is ugly. Perhaps deprecate `set` and write a new `setValue` function with a better argument order?

Enduriel:
I'm assuming this will be converted into the new kwarg format so will hold my review until then.

msu/systems/mod_settings/abstract_setting.nut
Enduriel: I think we go with the pain for now and add a new function that accepts kwargs

Enduriel (approved):

msu/systems/mod_settings/abstract_setting.nut
LordMidas: I mean eventually we have to do that anyway IF we do kwargs i.e. we'll have to add a new function. So this is fine.

LordMidas (approved):

PR #237, opened 2022-11-14
Simplify keybind settings by offloading the actual settings functionality to the modsettings system

TaroEld:
This depends on #235 .
This simplifies some aspects of the Keybinds system by making the Modsettings system entirely responsible for setting values, while they are only fetched directly from the setting. Persistent data is now also handled entirely over Modsettings. Concerns are further separated and the code is easier to read and maintain.

### DIFF
--- a/msu/systems/mod_settings/elements/keybind_setting.nut
+++ b/msu/systems/mod_settings/elements/keybind_setting.nut
@@ -1,10 +1,4 @@
 ::MSU.Class.KeybindSetting <- class extends ::MSU.Class.StringSetting
 {
 	static Type = "Keybind";
-	// Maybe add a keybind combination validator at some point
-
-	function printForParser()
-	{
-		base.printForParser(this.Type);
-	}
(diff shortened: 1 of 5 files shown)


msu/systems/mod_settings/elements/keybind_setting.nut
```diff
@@ -1,10 +1,4 @@
 ::MSU.Class.KeybindSetting <- class extends ::MSU.Class.StringSetting
 {
 	static Type = "Keybind";
-	// Maybe add a keybind combination validator at some point
```
Enduriel, line 4: I'd keep this line in just because it's a reminder for the future

Enduriel (approved):

msu/systems/mod_settings/elements/keybind_setting.nut
TaroEld: This has been added since, see #177

PR #238, opened 2022-11-16
refactor: add hard-coded names for vanilla DLC in popup errors

LordMidas:


Enduriel (approved):
This looks fine but
1) This should be a feat commit, this is a new functionality of translating dlc backend names to frontend pretty ones.
2) I don't like this style of switch statement with the double newline after the break. This isn't the only one like this in the codebase, I see Midas has been using them basically exclusively, so I won't request changes on this basis, but I think we should have a style commit to adjust all of these. The break is already effectively a newline for me since it doesn't really do anything you need to pay attention to

LordMidas:
> This looks fine but
> 
> 1. This should be a feat commit, this is a new functionality of translating dlc backend names to frontend pretty ones.
> 2. I don't like this style of switch statement with the double newline after the break. This isn't the only one like this in the codebase, I see Midas has been using them basically exclusively, so I won't request changes on this basis, but I think we should have a style commit to adjust all of these. The break is already effectively a newline for me since it doesn't really do anything you need to pay attention to

1. `feat` is a new feature though. I even considered `fix` label for this commit. This isn't a feature - it is simply a replacement of showing DLC name instead of id in the same feature that exists. I'd argue that a `fix` or `refactor` is more appropriate here.

Regarding 2: My personal preference is that a empty line after `break` makes it easier to see that there is a break, and if I don't have a `break` statement then I don't include an empty line which makes it very clear while reading that the switch is supposed to fall through to the next case. It's just more clear at a glance which cases fall through and which don't. But I can switch my style to yours if both you and @TaroEld prefer that.

msu/squirrel_hooks/mod_hooks.nut
```diff
@@ -154,6 +154,28 @@ local _mods_runQueue = ::_mods_runQueue;
 			{
 				if (dependencyMod == null)
 				{
+					switch (dependencyTable.Name)
```
TaroEld, line 157: A table with `dlcCode` : `dlcName` and an `in` check would probably look cleaner.

Enduriel:
> `feat` is a new feature though. I even considered `fix` label for this commit. This isn't a feature - it is simply a replacement of showing DLC name instead of id in the same feature that exists. I'd argue that a `fix` or `refactor `is more appropriate here.
The question is *why* you're doing this, are you changing something to organize it better, or because it will improve the experience of the user, if it is the latter, it cannot be a refactor. (which is why I think it should be a feat

LordMidas:
> > `feat` is a new feature though. I even considered `fix` label for this commit. This isn't a feature - it is simply a replacement of showing DLC name instead of id in the same feature that exists. I'd argue that a `fix` or `refactor `is more appropriate here.
> 
> The question is *why* you're doing this, are you changing something to organize it better, or because it will improve the experience of the user, if it is the latter, it cannot be a refactor. (which is why I think it should be a feat)

I disagree. Let's discuss this over Discord. All this commit does is print DLC name instead of id in popup errors. Improving the user's experience doesn't qualify something as a feat. A feat has to be a new addition to the feature set of the program. I know the name refactor doesn't really apply LITERALLY in this case, but that's how conventional commits uses it. Anything that isn't a feat or fix is a refactor.

msu/squirrel_hooks/mod_hooks.nut
LordMidas: This is just a patchy solution for the time being so I've kept it rather simple especially because all the keys and values are constants and don't need changing. For the future we should consider making a system where mod ids can be matched with mod names maybe via an online file or something.

PR #240, opened 2022-11-16
feat: kwargs arguments MSU.Table function

TaroEld:
This adds a function to get a key word arguments table that employs delegation. The goal is to allow for easier handling of large numbers of arguments where ordering might become cumbersome. Due to the delegation, `this` will first refer to the passed options, and then to the passed 'default' arguments.

### DIFF
--- a/scripts/config/msu/table.nut
+++ b/scripts/config/msu/table.nut
@@ -1,6 +1,6 @@
 ::MSU.Table <- {
 
-	function getKwargsTable(_options, _delegate)
+	function getKwargsTable( _options, _delegate )
 	{
 		if (_options == null)
 			_options = {};
--- a/scripts/config/msu/table.nut
+++ b/scripts/config/msu/table.nut
@@ -1,11 +1,11 @@
 ::MSU.Table <- {
 
-	function getKwargsTable( _options, _defaultArgumentTable )
+	function getKwargsTable( _passedArguments, _defaultArgumentsTable )
 	{
-		if (_options == null)
-			_options = {};
-		_options.setdelegate(_defaultArgumentTable);
-		return _options;
+		if (_passedArguments == null)
+			_passedArguments = {};
+		_passedArguments.setdelegate(_defaultArgumentsTable);
+		return _passedArguments;
 	}
 
 	function rand( _table )
--- a/scripts/config/msu/table.nut
+++ b/scripts/config/msu/table.nut
@@ -1,6 +1,6 @@
 ::MSU.Table <- {
 
-	function getKwargsTable( _passedArguments, _defaultArgumentsTable )
+	function kwargs( _passedArguments, _defaultArgumentsTable )
 	{
 		if (_passedArguments == null)
 			_passedArguments = {};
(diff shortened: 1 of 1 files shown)


Enduriel (approved):

LordMidas:
Could you please write up a use-case example of this?

LordMidas:
Based on discussion in Discord, we basically need a way to verify arg names and potentially allow modders to add more args.

LordMidas:
Also the `MSU.Table` namespace is specifically for working with squirrel tables (an indirect way of adding metamethods to `table` analogous to `array.` functions.). This `getKwargs` function should be part of the `MSU.` global namespace.

LordMidas:
Based on discussion in Discord we decided to go with something like:
```squirrel
function getKwargs( _passedArgs, _defaultArgs )
{
	if (_passedArgs = null) _passedArgs = {};
	foreach (key, value in _passedArgs)
	{
		if (!(key in _defaultArgs))
		{
			::logError(key + " is not a valid keyword argument for this function. If you are trying to add a new argument, you must hook the function and expand the default arguments table");
			throw ::MSU.Exception.InvalidValue(key);
		}
	}

	_passedArgs.setdelegate(_defaultArgs);
	return _passedArgs;
}
```

And then we enforce a standard that kwargs must be used as follows i.e. **with a null argument and an empty table argument**.

```squirrel
function myFunc( _options = null, _defaultArgs = {} )
{
	_defaultArgs.ArgName1 = false;
	_defaultArgs.ArgName2 = true;

	local args = ::MSU.getKwargs(_options, _defaultArgs);
}
```

And for hooking:

```squirrel
local someFunc = o.someFunc;
o.someFunc = function( _options = null, _defaultArgs = {} )
{
	_defaultArgs.MyAdditionalArg = 5;
	someFunc(_options, _defaultArgs);
}
```

LordMidas (changes requested):
1. Please move to `::MSU.` namespace.
2. I prefer the name `getKwargs` over `getKwargsTable`.

Based on discussion in Discord we decided to go with something like:
```squirrel
function getKwargs( _passedArgs, _defaultArgs )
{
	if (_passedArgs = null) _passedArgs = {};
	foreach (key, value in _passedArgs)
	{
		if (!(key in _defaultArgs))
		{
			::logError(key + " is not a valid keyword argument for this function. If you are trying to add a new argument, you must hook the function and expand the default arguments table");
			throw ::MSU.Exception.InvalidValue(key);
		}
	}

	_passedArgs.setdelegate(_defaultArgs);
	return _passedArgs;
}
```

And then we enforce a standard that kwargs must be used as follows i.e. **with a null argument and an empty table argument**.

```squirrel
function myFunc( _options = null, _defaultArgs = {} )
{
	_defaultArgs.ArgName1 = false;
	_defaultArgs.ArgName2 = true;

	local args = ::MSU.getKwargs(_options, _defaultArgs);
}
```

And for hooking:

```squirrel
local someFunc = o.someFunc;
o.someFunc = function( _options = null, _defaultArgs = {} )
{
	_defaultArgs.MyAdditionalArg = 5;
	someFunc(_options, _defaultArgs);
}
```

Enduriel:
We'll be taking a different approach with this after a lot of discussion so closing this PR in favor of a different one with the new approach.

PR #241, opened 2022-11-16
feat: print file, function & line with printData and debug print

Enduriel:
Pretty simple feature that I'd created right after we released 1.1.1 and somehow didn't creat a PR for


LordMidas:
Could you post a quick example of what it does as opposed to what it does right now?

TaroEld (approved):

LordMidas (approved):

PR #244, opened 2022-11-18
Ai behavior script only

LordMidas:
@Enduriel Please check if commits are properly labeled for this one :P

PR #245, opened 2022-11-18
Stack based behavior addition removal

LordMidas:
This is dependent on the changes from #244 and basically allows us to safely remove the behavior added by a skill if that skill is removed. It is far simpler than the skills thing because we don't have to care about any references or items or anything.

### DIFF
--- a/msu/hooks/ai/tactical/agent.nut
+++ b/msu/hooks/ai/tactical/agent.nut
@@ -1,0 +1,33 @@
+::mods_hookBaseClass("ai/tactical/agent", function(o) {
+	o = o[o.SuperName];
+
+	o.m.MSU_BehaviorStacks <- {};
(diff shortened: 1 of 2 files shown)


Enduriel:
Since this is dependent on the changes for #244, please either merge in #244 or retarget this to merge into ai-behavior-script-only

LordMidas:
@Enduriel and @TaroEld review please

Enduriel (approved):

TaroEld (approved):

Enduriel (approved):
approved dependent on removeBehaviorIncrementally being put up for renaming at some point in the future

Enduriel (approved):
approved dependent on removeBehaviorByStack being put up for renaming at some point in the future

PR #247, opened 2022-11-19
feat: add MSU.Class.Regexp class

LordMidas:
The class expands regex functionality with additional functions. Please suggest better function names and/or function organization.

### DIFF
--- a/scripts/config/msu/classes/regexp.nut
+++ b/scripts/config/msu/classes/regexp.nut
@@ -1,0 +1,62 @@
+::MSU.Class.Regexp <- class
+{
+	__regexp = null;
+
+	constructor( _expression )
+	{
+		::MSU.requireString(_expression);
+		this.__regexp = regexp(_expression);
+	}
+
+	function search( _string, _startIndex = 0 )
+	{
+		return this.__regexp.search(_string, _startIndex);
+	}
+
+	function capture( _string, _startIndex = 0 )
+	{
+		return this.__regexp.capture(_string, _startIndex);
+	}
+
+	function match( _string )
+	{
+		return this.__regexp.match(_string);
+	}
+
+	function captureAll( _string, _startIndex = 0 )
+	{
+		local captures = [];
+		while (true)
+		{
+			local capture = this.capture(_string, _startIndex);
+			if (capture == null) break;
+
+			captures.push(capture);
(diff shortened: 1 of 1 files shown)


scripts/config/msu/classes/regexp.nut
```diff
@@ -0,0 +1,52 @@
...
+	constructor( _expression )
+	{
+		::MSU.requireString(_expression);
+		this.__regexp = regexp(_expression);
+	}
+
+	function _get( _key )
+	{
+		return this.__regexp[_key];
+	}
```
Enduriel, line 14: You can't just overwrite `_get` like this, if you do you will lose this class when you call stuff like `.weakref()` or `.getclass()`. I'm not sure if there is a better solution than just having an explicit `throw null` for those scenarios, but you should probably handle those two at a minimum

Enduriel (changes requested):

scripts/config/msu/classes/regexp.nut
LordMidas: Based on my testing using `.getclass()` returns the correct class i.e. `::MSU.Class.Regexp`, so that works already. Using `weakref` returns a weak reference but I am unable to call any functions on it (neither regular regexp functions e.g. `capture` or our class functions e.g. `captureAll`). I've tried `if (_key == "weakref") throw null` in `_get` but it doesn't help. So we could implement:
```squirrel
function weakref()
{
   throw "The MSU Regexp class does not support weakref";
}
```

LordMidas: This is not true. Based on my testing both `weakref()` and `getclass()` return a weak ref to the Regexp class instance and the Regexp class respectively. Both assertions pass:
```
local reg = ::MSU.Class.Regexp("dfdfd");
local weak = reg.weakref();
local c = reg.getclass();
assert(c == ::MSU.Class.Regexp);
assert(weak.ref() == reg);
```
However, I am unable to call any functions on the weakref e.g. `weak.capture` and `weak.captureAll` both throw index not found exceptions. However calling `reg.capture` and `reg.captureAll` works properly. The problem persists even if I add `if (_key == "weakref") throw null` in the `_get` metamethod. Any thoughts on why that is happening?

PR #248, opened 2022-11-22
feat: return first passed table of ::MSU.Table.merge()

TaroEld:
This makes it more convenient to use, for example with clone().


Enduriel (approved):

LordMidas (approved):

LordMidas:
This should be a `feat` commit.

PR #250, opened 2022-11-30
feat: add early JS registration

Enduriel:
Allows JS scripts to load before vanilla screens (except RootScreen) are instantiated, meaning we can now freely modify stuff like eventListeners or modules.
I don't really like the way this is organized, am open to suggestions


scripts/!mods_preload/msu.nut
```diff
@@ -22,6 +22,37 @@ local _mods_runQueue = ::_mods_runQueue;
 	::MSU.EndQueue.run();
 }
 
+::MSU.EndQueue.add(function()
+{
+	::mods_hookExactClass("root_state", function (o)
+	{
+		local onInit = o.onInit;
```
TaroEld, line 29: This line should be above the onInit overwrite (l37)

TaroEld (approved):
Works I suppose, but at some point wee should probably talk about just overwriting some of the more crucial files, as solutions like this are extremely hacky.

Enduriel:
Closed since this is being handled by Modern Hooks

PR #251, opened 2022-12-05
feat: add modsetting that supresses the basegame keybinds so that onl…

TaroEld:
…y the MSU system is used
The point is to be able to properly unbind and rebind keys, without the base keybinds still being active.

### DIFF
--- a/msu/hooks/states/main_menu_state.nut
+++ b/msu/hooks/states/main_menu_state.nut
@@ -77,7 +77,7 @@
 		{
 			return onKeyInput(_key);
 		}
-		if (::MSU.System.Keybinds.onKeyInput(_key, this, ::MSU.Key.State.MainMenu) || ::MSU.Mod.ModSettings.getSetting("suppressBaseKeybinds").getValue())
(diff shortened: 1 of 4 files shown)


LordMidas (changes requested):
The correct spelling is "Suppress" with a double p. Please correct throughout the PR in variable names and description tooltip. 

On a side note, perhaps we should change the name of the "Vanilla" panel in keybinds to "Vanilla (MSU)" to make it clear that these are the MSU keybinds for vanilla. I have recently seen a user for whom this wasn't clear.

LordMidas (approved):
Minor if you feel like changing now:
Better wording for tooltip: "Whether base keybinds should be suppressed. This means that only the MSU system will be used for keybinds.\nFor example, if you set 'Open Character Screen' from 'c' to 'tab' then without this setting, pressing 'c' will still open the character screen. With this setting, only 'tab' will open it.");

Enduriel:
With the caveat that I assume the ID will get capitalized in another PR

TaroEld:
> 

No reason I can't do it in this one

PR #253, opened 2022-12-12
feat: use behavior ID only

LordMidas:
This is accomplished via a hook on root_state onInit function that populates an ID to Script file name map. This way we don't need the user to pass both the id and the script file name for the behavior.

### DIFF
--- a/scripts/config/msu/ai.nut
+++ b/scripts/config/msu/ai.nut
@@ -1,4 +1,21 @@
 ::MSU.AI <- {
+	BehaviorIDToScriptMap = {}, // Is populated during root_state.onInit function
+
+	function getBehaviorScriptFromID( _id )
+	{
+		if (this.m.BehaviorIDToScriptMap.len() == 0)
+		{
+			throw "This function cannot be used before the root_state.onInit function has run");
+		}
+		if (!(_id in this.m.BehaviorIDToScriptMap))
+		{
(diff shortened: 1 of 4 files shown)


TaroEld (approved):

scripts/config/msu/ai.nut
```diff
@@ -1,4 +1,21 @@
...
+
+	function getBehaviorScriptFromID( _id )
+	{
+		if (this.m.BehaviorIDToScriptMap.len() == 0)
+		{
+			throw "This function cannot be used before the root_state.onInit function has run";
+		}
+		if (!(_id in this.m.BehaviorIDToScriptMap))
+		{
+			::logError("Invalid behavior id '" + _id + "' and/or the associated script file is not located in \'scripts/ai/tactical/behaviors\'");
```
Enduriel, line 12: unescaped `'` around _id, but escaped `\'` around `scripts/ai/tactical/behaviors`. Not certain which one is the right one (pretty sure it's the first though), but we definitely shouldn't have 2 different ones.

Enduriel (approved):
Other than that small thing lgtm

scripts/config/msu/ai.nut
LordMidas: In my experience squirrel doesn't care but I still use the escape characters *shrug*. We can decide what convention we want to follow.

LordMidas: We decided to go with escape characters as that is what the decompiler does for single quotes.

PR #258, opened 2022-12-27
feat: add onSkillsUpdated event

LordMidas:
This implementation is superior to #208 as it accomplishes the event without the requirement of a flag variable in `m`. It also fixes an issue with the other implementation by ensuring that the update keeps looping until all skills have been added / removed. Without this fix there was a possibility that skills added/removed in onSkillsUpdated events based on changes in the previous call to the event would not be properly added/removed.

### DIFF
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -219,6 +219,10 @@
 		setFatigueCost(_f);
 	}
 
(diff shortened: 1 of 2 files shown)


LordMidas:
Unnecessary, because tile tooltip isn't even shown when there is an actor on the target tile - it shows the Hit Factors tooltip instead.

PR #259, opened 2022-12-27
feat: add _asTarget parameter to onQueryTileTooltip

LordMidas:
This allows the skills of the entity on the target tile to also affect the tooltip. What are your thoughts about the `_tile.getEntity().getID() != this.getContainer().getActor().getID()` condition? I kept it in for consistency with `onGetHitFactors` and `onGetHitFactorsAsTarget` where the latter is only called if the target actor is different from the user. (Perhaps we should revise that too and call it even if the user is targeting himself)?

### DIFF
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -272,7 +272,7 @@
 	{
 	}
 
-	o.onQueryTileTooltip <- function( _tile, _tooltip )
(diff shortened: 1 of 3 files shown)


LordMidas:
Unnecessary as the tile tooltip isn't even shown when there is an actor on the target tile - the Hit Factors tooltip is shown instead.

PR #260, opened 2022-12-27
Segregate deserialize functions scope

LordMidas:
This `Utils.deserialize` function should not change the values of any object - as this violates the single responsibility principle. It should take a metadata object and deserialize an object from it and return the deserialized object.


Enduriel (approved):
I'm still not a huge fan of the merge being its own step, but since this is mostly a performance implication, I can live with it.

TaroEld (approved):

PR #261, opened 2022-12-28
feat: add Array.merge function

LordMidas:


Enduriel (changes requested):
remove recursively parameter as it can cause strange behaviors in my opinion and isn't really necessary for arrays

LordMidas:
I honestly can't remember the use case for this anymore, so I'd suggest we put this in postponed for now.

PR #264, opened 2022-12-30
fix: always passing true instead of _overwrite during recursive merge

LordMidas:


TaroEld (approved):

Enduriel (approved):

Enduriel:
This is redundant though since the code will never get to that point unless _overwrite is true

PR #265, opened 2022-12-31
feat: add functions to get neighboring tiles

LordMidas:
The utility of the `getNeighbors` function is obvious. The utility of the `getNeighbor` function is shown below:
```squirrel
// Objective:
// Get the first empty tile adjacent to myTile

// current workflow
local targetTile;
for (local i = 0; i < 6; i++)
{
	if (myTile.hasNextTile(i))
	{
		local nextTile = myTile.getNextTile(i);
		if (nextTile.IsEmpty)
		{
			targetTile = nextTile;
			break;
		}
	}
}

// new workflow:
local targetTile = ::MSU.Tile.getNeighbor(myTile, @(tile) tile.IsEmpty);
```

### DIFF
--- a/msu/utils/tile.nut
+++ b/msu/utils/tile.nut
@@ -15,7 +15,9 @@
 
 	function getNeighbors( _tile, _function = null )
 	{
-		// _function( tile ) and returns boolean
+		// _function( tile ) returns boolean
+		// iterates over all neighboring tiles and calls _function on each tile
+		// returns an array of tiles for which _function returned true
 
 		local ret = [];
 		for (local i = 0; i < 6; i++)
@@ -32,7 +34,9 @@
 
 	function getNeighbor( _tile, _function = null )
 	{
-		// _function( tile ) and returns boolean
+		// _function( tile ) returns boolean
+		// iterates over all neighboring tiles and calls _function on each tile
+		// returns the first tile for which _function returned true
 
 		for (local i = 0; i < 6; i++)
 		{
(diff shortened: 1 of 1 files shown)


msu/utils/tile.nut
```diff
@@ -12,4 +12,29 @@
...
+		{
+			if (_tile.hasNextTile(i))
+				ret.push(_tile.getNextTile(i));
+		}
+		return ret;
+	}
+
+	function getNeighbor( _tile, _function )
+	{
+		// _function( tile ) and returns boolean
```
Enduriel, line 29: this needs to be better, doesn't really explain anything and should be using `_tile` regardless. (can be removed though and that would be fine imo.

Enduriel (changes requested):

msu/utils/tile.nut
LordMidas: The function is correct. It does the following: "Give me the first neighbor of `_tile` for which `_function` returns true`.

Enduriel: I'm not saying the function is wrong, I'm saying the comment isn't useful

Enduriel (changes requested):
The comments here are still not very useful and should be changed or removed

LordMidas:
> The comments here are still not very useful and should be changed or removed

I don't know why you say that. Looking at the comment it becomes clear what the format of the passed `_function` should be. And looking at the code's body it becomes clear how that function's return value is used. Why does the comment need to be any more elaborate than that? The code is self-explanatory.

Enduriel (approved):

TaroEld (approved):

PR #266, opened 2023-01-02
feat: add isDisarmed function

LordMidas:
In vanilla there is no good way to check if a character is disarmed except by checking for `effects.disarmed` effect. However, if a mod adds another way to apply the disarmed status to a character, then other mods won't be able to check for that. This function allows a centralized isDisarmed function that mods can hook to add their checks into it.


Enduriel (approved):

TaroEld (approved):
Not sure how needed this is but ok

PR #269, opened 2023-01-04
feat: no longer prints empty variables lines when using printStackTrace

Enduriel:
also restructured the function to be a little cleaner.

This is to put it in line with the way vanilla prints stack traces to a log.


TaroEld (approved):

LordMidas (approved):
I don't understand each line of code here (almost 2 AM) but I get the overall idea so I approve because it werks (I believe).

PR #271, opened 2023-01-06
feat: write text instead of operator for required version

LordMidas:
This helps people who are unable to understand math comparison operators.

### DIFF
--- a/msu/squirrel_hooks/mod_hooks.nut
+++ b/msu/squirrel_hooks/mod_hooks.nut
@@ -122,28 +122,28 @@ local _mods_runQueue = ::_mods_runQueue;
 		return compareSemVerVersionsWithNulls(_modTable.SemVer, _dependencyTable.VersionOperator, _dependencyTable.SemVer);
 	}
 
-	local function getTextForOperator( _operator )
+	local function getOperatorVersionAsText( _operator, _version )
 	{
 		switch (_operator)
 		{
 			case ">":
-				return "newer than";
+				return "newer than " + _version;
 
 			case ">=":
-				return "at least";
+				return _version + " or newer";
 
 			case null:
 			case "=":
-				return "";
+				return _version;
 
 			case "!=":
-				return "other than";
+				return "other than " + _version;
 
 			case "<=":
-				return "up to";
+				return _version + " or older";
 
 			case "<":
-				return "older than";
+				return "older than " _version;
 
 			default:
 				throw ::MSU.Exception.InvalidValue(_operator);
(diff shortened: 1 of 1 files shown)


msu/squirrel_hooks/mod_hooks.nut
```diff
@@ -140,7 +168,7 @@ local _mods_runQueue = ::_mods_runQueue;
 					else if (compareVersions(dependencyTable, dependencyMod))
 					{
 						local tableVersion = dependencyTable.SemVer == null ? dependencyTable.Version : ::MSU.SemVer.getVersionString(dependencyTable.SemVer);
-						::MSU.QueueErrors.add(mod.FriendlyName + " is not compatible with " + dependencyMod.FriendlyName + " version " + dependencyTable.VersionOperator + tableVersion);
+						::MSU.QueueErrors.add(mod.FriendlyName + " is not compatible with " + dependencyMod.FriendlyName + " version " + getTextForOperator(dependencyTable.VersionOperator) + " " + tableVersion);
```
Enduriel, line 171: Make version plural so it's `versions`, that makes more sense if we're structuring the sentence this way now.
eg:
`Mod X is not compatible with Mod Y versions older than 1.0.0`
`Mod A is not compatible with Mod Y versions up to 2.0.0`

msu/squirrel_hooks/mod_hooks.nut
```diff
@@ -194,7 +222,7 @@ local _mods_runQueue = ::_mods_runQueue;
 			{
 				local tableVersion = dependencyTable.SemVer == null ? dependencyTable.Version : ::MSU.SemVer.getVersionString(dependencyTable.SemVer);
 				local modVersion = dependencyMod.SemVer == null ? dependencyMod.Version : ::MSU.SemVer.getVersionString(dependencyMod.SemVer);
-				::MSU.QueueErrors.add(mod.FriendlyName + " requires " + dependencyMod.FriendlyName + " version " + dependencyTable.VersionOperator + tableVersion + " but version " + modVersion + " was found");
+				::MSU.QueueErrors.add(mod.FriendlyName + " requires " + dependencyMod.FriendlyName + " version " + getTextForOperator(dependencyTable.VersionOperator) + " " + tableVersion + " but version " + modVersion + " was found");
```
Enduriel, line 225: as above

Enduriel (changes requested):

msu/squirrel_hooks/mod_hooks.nut
LordMidas: But that will not work with situations where it says `=` or `!=` e.g.
`Mod x is not compatible with Mod Y versions 1.2.0` - makes no sense. So I'd have to wrap the word "version" within the `getTextForOperator` function and return singular/plural for different cases.

PR #274, opened 2023-01-08
fix: clearFlags getting called before most deserialization

Enduriel:
This causes issues where loading a save two times without loading makes the save not load properly due to clearFlags clearing out all the flags that hold data. With this fix clearFlags will always be called after any `onDeserialize` functions. For the sake of completeness, this has been mirrored on the serialization side.

This also adds a new pair of hookable functions which probably should be documented at some point, but this seems inappropriate for an RC. I am very open to suggests for how to improve the function names here.


TaroEld (approved):

Enduriel:
This is superseded by #280

PR #275, opened 2023-01-11
fix(vanilla): call onCombatStarted also for ai characters

LordMidas:


msu/hooks/entity/tactical/tactical_entity_manager.nut
```diff
@@ -1,4 +1,27 @@
...
+				foreach (actor in faction)
+				{
+					actor.getSkills().onCombatStarted();
+					actor.getItems().onCombatStarted();
+					actor.getSkills().update();
+				}
+			}
+		}
+
+		::Math.seedRandom(::Time.getRealTime());
```
Enduriel, line 20: Not sure why we need to do this since the end of the spawn function already takes care of re-randomizing the seed in vanilla.

Enduriel (changes requested):

msu/hooks/entity/tactical/tactical_entity_manager.nut
LordMidas: I didn't know what was the purpose of that line at the end of the vanilla function, so I just added it here to be safe in case its relevant for something. Should I remove it?

Enduriel (approved):

LordMidas:
I still think we should do this by adding `onCombatStart` to `actor.nut` i.e. the original solution I proposed in my bug report to vanilla. I recently had a situation in Reforged where I would have liked the existence of such a function for AI characters.

Enduriel:
tbh I'm not sure why I was opposed to that before but that doesn't sound like an unreasonable addition to me. If we do go with that we should make sure we hook player to also call the parent function.

Enduriel:
I took a look and I'm honestly okay with overwriting that, I double checked legends and even they don't change anything here, and tbh I don't see anyone needing to change this except to extend it in which case they should hook.

Enduriel (approved):

msu/hooks/entity/tactical/humans/standard_bearer.nut
```diff
@@ -0,0 +1,11 @@
+// Is included during First QueueBucket
+// VANILLAFIX: http://battlebrothersgame.com/forums/topic/oncombatstarted-is-not-called-for-ai-characters/
+// This fix is spread out over 4 files: tactical_entity_manager, actor, player, standard_bearer
+::MSU.HooksMod.hook("scripts/entity/tactical/humans/standard_bearer", function(q) {
+	q.onInit = @(__original) function()
+	{
+		__original();
+		this.m.Skills.removeByID("perk.inspiring_presence");
+		::logWarning("MSU removed perk_inspiring_presence from standard_bearer. We do this to maintain unchanged vanilla behavior after our fix of onCombatStart for AI characters").
```
TaroEld, line 9: Remove that warning though.

TaroEld (approved):

msu/hooks/entity/tactical/humans/standard_bearer.nut
LordMidas: The warning is there so that other modders using MSU are aware of this change as this is a gameplay change happening in MSU. Otherwise modders who expect that perk on standard_bearer and make changes to the perk will be surprised to see that the perk isn't present on those enemies.

TaroEld: Okay, seems like would need to add a bunch more warnings then to account for all the stuff that we changed or fixed and unaware modders might not expect?

TaroEld (approved):

msu/hooks/entity/tactical/humans/standard_bearer.nut
LordMidas: This is different from other changes though. If we don't remove the `perk_inspiring_presence` from the standard_bearer that changes vanilla behavior. But changing perk setup of entities seems like something that is outside the scope of MSU so it seems like this is a kind of change no one expects MSU to make - which is why I put an explicit warning for it. Idk what do you think @Enduriel ?

PR #276, opened 2023-01-17
feat: re implement onMovementStep to be able to affect actor movement

LordMidas:
This makes the function actually useful. It allows skills to stop an actor's movement at a particular tile based on conditions defined in the skill's `onMovementStep` function. Note that `onMovementStep` is called for all "to-be-moved-on" tiles even before the movement starts. As long as all skills' `onMovementStep` returns true, the movement is valid. If even a single skill returns false, the movement is stopped at that tile. If any skill returns null, then the movement is stopped and the movement cost refunded.

### DIFF
--- a/msu/hooks/skills/skill_container.nut
+++ b/msu/hooks/skills/skill_container.nut
@@ -96,10 +96,44 @@
 
 	q.onMovementStep <- function( _tile, _levelDifference )
 	{
-		this.callSkillsFunction("onMovementStep", [
-			_tile,
-			_levelDifference
-		], false);
+		local wasUpdating = this.m.IsUpdating;
+		this.m.IsUpdating = true;
+		this.m.IsBusy = false;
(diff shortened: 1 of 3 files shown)


msu/hooks/skills/skill_container.nut
```diff
@@ -98,10 +98,44 @@
...
+						break;
+
+					case null:
+						canMove = false;
+						isRefunding = true;
+						break;
+
+					default:
+						::logError("onMovementStep must return true, false, or null");
+						throw "invalid return value";
```
TaroEld, line 129: Well this aint it. Rather have an Enum with the 3 states that can be returned. Or pass in a table where values can be changed by the function.

LordMidas: Yeah this will be changed. @Enduriel wants something with partial refunds too, so it'll probably have to return a table or something.

PR #280, opened 2023-01-25
fix: flagSerialization edge case errors

Enduriel:
serialization of BB objects is now handled by passing a Serialization or Deserialization Emulator to them.

Normal deserialiation now requires a default value.

SerDeEmulators are now permanently linked to a FlagContainer they are assigned during instantiation.

SerializationEmulators now have an Incremental mode where they store their data in the flag container as soon as something is written to them.

DeserializationEmulators now immediately clear flags from their FlagContainer as they deserialize.

SerDeEmulator flag IDs are validated to make sure they do not collide with flags generated by other emulators.

### DIFF
--- a/msu/systems/mod_settings/abstract_setting.nut
+++ b/msu/systems/mod_settings/abstract_setting.nut
@@ -186,16 +186,21 @@
 		if (_table.Locked) this.lock(_table.LockReason);
 	}
 
+	function getSerDeFlag()
+	{
+		return "MS." + this.getID();
+	}
+
 	function flagSerialize( _out )
 	{
-		this.getMod().Serialization.flagSerialize(format("MS.%s", this.getID()), this.__getSerializationTable());
+		this.getMod().Serialization.flagSerialize(this.getSerDeFlag(), this.__getSerializationTable());
 	}
 
 	function flagDeserialize( _in )
 	{
 		if (::MSU.Mod.Serialization.isSavedVersionAtLeast("1.2.0-rc.1", _in.getMetaData()))
 		{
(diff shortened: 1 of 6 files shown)


msu/systems/mod_settings/abstract_setting.nut
```diff
@@ -195,7 +195,7 @@
 	{
 		if (::MSU.Mod.Serialization.isSavedVersionAtLeast("1.2.0-rc.1", _in.getMetaData()))
 		{
-			this.__setFromSerializationTable(this.getMod().Serialization.flagDeserialize(format("MS.%s", this.getID())));
+			this.__setFromSerializationTable(this.getMod().Serialization.flagDeserialize(format("MS.%s", this.getID()), this.__getSerializationTable()));
```
TaroEld: I'd prefer if we defined the format string either the line before or via a separate function so this line becomes more readable

msu/systems/serialization/deserialization_emulator.nut
```diff
@@ -68,17 +68,21 @@
 		return this.__readFloat();
 	}
 
-	function loadDataFromFlagContainer( _flags ) // doesn't check if flags exist
+	// function hasFlag()
+	// {
+	// 	return this.FlagContainer.has(this.getEmulatorString());
+	// }
```
TaroEld: We talked about removing this until there's a need for it

TaroEld (approved):

LordMidas (changes requested):
@Enduriel When you have time please explain to me the EmulatorsToClear and Incremental thingies.

TaroEld (approved):

LordMidas (approved):
Please replace `MS.` with `ModSetting.` and then we're good to go.

PR #281, opened 2023-01-27
feat: set the AIBehaviorID of skills to vanilla behavior during create

LordMidas:
Is accomplished via a hookDescendants on skill.nut which looks for the `create` function and then adds a custom function which is called at the end of the hooked `create`. The function uses a table which is populated during root_state.onInit function to match the skill id to its vanilla behavior id. This is done to avoid having to instantiate any file during queue, which breaks hooks.

### DIFF
--- a/msu/hooks/root_state.nut
+++ b/msu/hooks/root_state.nut
@@ -6,7 +6,16 @@
 		{
 			try
 			{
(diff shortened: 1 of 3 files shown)


Enduriel:
Since we will be taking the hardcoded approach in #282, this PR is no longer necessary.

PR #282, opened 2023-01-27
feat: set the AIBehaviorID of skills to vanilla behavior during create

LordMidas:
This is accomplished via a manual list of all the skills mapped to their behavior ids as they are present in vanilla as of version 1.5.0.13


Enduriel (approved):

TaroEld (approved):
We might want to move the map into a different file at some point because it's big, but good for now

PR #278, opened 2023-01-20
fix: check required setting values during mod settings deserialization

LordMidas:

PR #283, opened 2023-02-01
Individual locks for settings

LordMidas:
- Allows individual locks to be placed on settings which can be removed on a per-lock basis.
- Major advantage: this allows settings, during their on-changed callbacks, to add/remove locks on other settings. 
- A setting is only considered unlocked once all the locks have been removed.
- While the `lock` function is deprecated, backwards compatible support for it is maintained via a legacy lock id.
- The Locked state is removed from a setting's serialization, because locks should be set via `requireSettingValue` and/or `addLock`. Serializing the locked state leads to unintended behavior when mods are added/removed.

Request: We need to add functionality, so that after "Saving" and/or "Applying" the settings - we run the requiredSettingValues check (just like we do on campaign load - see #278). And if any setting couldn't be changed to a value required by the on-change callbacks or now has a value that violates required setting values, we should throw a popup/notification to notify the user.

### DIFF
--- a/msu/systems/mod_settings/abstract_setting.nut
+++ b/msu/systems/mod_settings/abstract_setting.nut
@@ -3,8 +3,7 @@
 	static Type = "Abstract";
 	Value = null;
 	BaseValue = null;
-	Locked = null;
-	LockReason = null;
+	Locks = null;
 	BeforeChangeCallbacks = null;
 	AfterChangeCallbacks = null;
 	Persistence = null; //if it should print change to log for further manipulation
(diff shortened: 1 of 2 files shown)


LordMidas:
@Suor If you are interested in mod settings currently and their "locking" behavior, this very old PR of mine might be interesting for you. It probably needs to be reworked as it is very old now, but the concept seems important/relevant. Let me know your thoughts.

Suor:
The idea makes sense but this is a complication. I would make some estimate how many people actually need this.

PR #288, opened 2023-02-12
fix: remove check for existing behavior

LordMidas:
This is important because we add behavior by stack. And if we check for pre-existing behavior, then we don't add the behavior but we still remove it when the skill is removed. This leads to issues when a skill removes itself upon execution e.g. Throw Spear or Throw Net which get removed due to the item being one-use only. As this happens mid-execution of an ai behavior, the behavior is removed and the game freezing as the behavior is no longer attached to an actor.

### DIFF
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -208,7 +208,7 @@
 
 		setContainer(_c);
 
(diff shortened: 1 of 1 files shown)


TaroEld (approved):
Makes sense to me

Enduriel (approved):

PR #292, opened 2023-02-23
fix: add check to avoid adding the same item twice

LordMidas:


LordMidas:
Bad branch. Will rebase and re-open PR.

PR #293, opened 2023-02-23
fix: add check to avoid adding the same item twice

LordMidas:


TaroEld (approved):
Seems to work fine this way

Enduriel (approved):

PR #297, opened 2023-03-08
fix: metadata emulators being empty and potentially causing issues

Enduriel:
this would occur if a mod tried to use a property of metadata they expected to exist.
For example if they use a Serialization.isSavedVersionAtLeast() check on the emulated metadata this would error.
This method extract the correct information from world_state, however it has obvious issues in that this is very non-standard behavior

### DIFF
--- a/msu/systems/serialization/load.nut
+++ b/msu/systems/serialization/load.nut
@@ -10,3 +10,19 @@ includeFile("deserialization_emulator");
 includeFile("serialization_system.nut");
 ::MSU.System.Serialization <- ::MSU.Class.SerializationSystem();
 includeFile("serialization_mod_addon.nut");
+
+::MSU.UI.addOnConnectCallback(function()
+{
+	local worldState = ::new("scripts/states/world_state");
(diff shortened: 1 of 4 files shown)


TaroEld (approved):
This is terrible but I don't really have an alternative beyond equally terrible options

msu/systems/serialization/load.nut
```diff
@@ -10,3 +10,19 @@ includeFile("deserialization_emulator");
 includeFile("serialization_system.nut");
 ::MSU.System.Serialization <- ::MSU.Class.SerializationSystem();
 includeFile("serialization_mod_addon.nut");
+
+::MSU.UI.addOnConnectCallback(function()
+{
+	local worldState = ::new("scripts/states/world_state");
+	local serEm = ::MSU.Class.SerializationEmulator(::MSU.ID, "WorldStateOnBeforeSerialize", ::new("scripts/tools/tag_collection"), ::MSU.Class.MetaDataEmulator());
+	::MSU.System.Serialization.MetaData = serEm.getMetaData();
+	::World.Assets <- {
```
LordMidas, line 19: Please add a `_get` metamethod for missing entries.

LordMidas (approved):

PR #304, opened 2023-04-13
refactor: remove automatic AI behavior adding 

TaroEld:
These changes came with some unintended vanilla changes, where units got behaviors they were not supposed to have (eg zombies got knockout and such).


LordMidas (approved):

TaroEld:
Added directly to main and released in 1.2.5

PR #313, opened 2023-09-12
refactor: set costs preview before calling skills function

LordMidas:
This ensures that the actor's PreviewFatigue and PreviewActionPoints fields are properly set BEFORE the skill_container's onAffordabilityPreview function is called. This is currently a problem because those fields are set AFTER the event -  so the values during the event call are wrong.

However, this means that we need to update the UI again after calling onAffordabilityPreview (as the UI is initially updated in the call to the base setActiveEntityCostsPreview function). We could just call the `updateCostsPreview` on the JSHandle again manually, however that leads to somewhat glitchy animation in the UI as it tries to update twice. 

Therefore, we implement a switch to disable the UI updating during the call to the base function, and then enable it after calling our onAffordabilityPreview event, after which we manually call the updateCostsPreview for the JSHandle update.


Enduriel (changes requested):
As an alternative to having 3 extra JS calls, I would recommend an advanced switcheroo on asyncCall with the following approach:

```squirrel
local table = {
	__JSHandle = this.m.JSHandle,
	function asyncCall( _funcName, ... )
	{
		if (_funcName == "updateCostsPreview")
			return;
		vargv.insert(0, this);
		vargv.insert(1, _funcName);
		this.__JSHandle.asyncCall.acall(vargv);
	}
}
table.setdelegate({
	function _get( _key )
	{
		return this.__JSHandle[_key];
	}
})
this.m.JSHandle = table;
__original(_costsPreview);
this.m.JSHandle = table.__JSHandle;
```

LordMidas:
I've pushed the changes with the switcheroo. However, this branch should wait until the port to modern hooks is first merged, then I'll test it before merging.

msu/ui/load.nut
```diff
@@ -6,6 +6,7 @@
 ::mods_registerJS("msu/ui_hooks/main_menu_module.js");
 ::mods_registerJS("msu/ui_hooks/main_menu_screen.js");
 ::mods_registerJS("msu/ui_hooks/tooltip_module.js");
+::mods_registerJS("msu/ui_hooks/turnsequencebar_module.js");
```
Enduriel, line 10: This can now be removed

Enduriel (changes requested):

TaroEld:
I don't know what the previous thing was but I rather have three js calls than getting to clever with these switches and perhaps inadvertently breaking something.

LordMidas:
> I don't know what the previous thing was but I rather have three js calls than getting to clever with these switches and perhaps inadvertently breaking something.

See the commit history of the PR. Specifically this commit: https://github.com/MSUTeam/MSU/pull/313/commits/c67045111962a1247b96b5bc33ec28417de161c7

TaroEld:
It's not very pretty but I still prefer it.

LordMidas:
@Enduriel , @TaroEld decide between yourselves which approach you prefer i.e. 3 JS calls or the switheroo method. I'll do whatever the two of you decide.

msu/hooks/ui/screens/tactical/modules/turn_sequence_bar/turn_sequence_bar.nut
```diff
@@ -28,7 +28,29 @@
...
+				this.__JSHandle.asyncCall.acall(vargv);
+			}
+		}
+		table.setdelegate({
+			function _get( _key )
+			{
+				if (_key in this.__JSHandle) return this.__JSHandle[_key];
+				throw null;
+			}
+		});
```
TaroEld, line 49: Probably define it in .m once and then refer to it instead of redefining it every time.

TaroEld (approved):

msu/hooks/ui/screens/tactical/modules/turn_sequence_bar/turn_sequence_bar.nut
LordMidas: Done

TaroEld (approved):

msu/hooks/ui/screens/tactical/modules/turn_sequence_bar/turn_sequence_bar.nut
```diff
@@ -1,4 +1,28 @@
...
+
+				vargv.insert(0, _funcName);
+				vargv.insert(0, this);
+				this.__JSHandle.asyncCall.acall(vargv);
+			}
+		}
+		this.m.MSU_JSHandle.setdelegate({
+			function _get( _key )
+			{
+				if (_key in this.__JSHandle) return this.__JSHandle[_key];
```
Enduriel, line 20: newline the return

Enduriel (approved):
After the one syntactical change it lgtm

PR #315, opened 2023-09-13
feat: add utility function for adding info for new entities

LordMidas:


msu/utils/entities.nut
```diff
@@ -0,0 +1,32 @@
+::MSU.Entities <- {
+	EntityTypeToDefaultFactionMap = {},
+	HighestIDDiff = -99,
+	HighestID = -99,
```
Enduriel, line 4: why -99? might as well just make it -1

Enduriel (changes requested):

msu/utils/entities.nut
LordMidas: `-1` is the ID of `::Const.EntityType.Player`. And I don't want to do `-2`. So I went with an obvious negative number that shows intent by going `-99`. Open to other suggestions.

TaroEld: >-1 is the ID of ::Const.EntityType.Player
oh my lol

Darxo:
Finding out the highest ID for every single new entity is not very efficient.
I suggest a second function for adding an array of EntityInfo which only calculates the highest ID once.

Something like in this PR over at Reforged: https://github.com/Battle-Modders/mod-reforged/pull/247/files#

image

msu/utils/entities.nut
```diff
@@ -0,0 +1,32 @@
...
+	function addEntity( _info )
+	{
+		// _info is a table that must contain the following keys:
+		// EntityType, EntityIcon, DefaultFaction, EntityName, EntityNamePlural, TroopDef, ActorDef
+
+		// If the Diff is now greater than before, that means some other mod added something to the ::Const.EntityType
+		// table, so we should update our HighestID again.
+		if (this.HighestID == -99 || ::Const.EntityType.len() - this.HighestID > this.HighestIDDiff)
+		{
+			foreach (key, value in ::Const.EntityType)
```
Darxo, line 16: if you don't use `key` you should not include it here

msu/utils/entities.nut
Enduriel: wow pain

LordMidas:
Upon gaining more experience with adding new entities, I have gained more insight that leads to me to think that this PR is not enough. Namely `EntityType` is only a thing to count entities and get their names in vanilla. So we need to handle them separately from `Troops` which are actual units that spawn on the map. So I'll think more on this PR and come up with better solutions.

Enduriel:
In that case let's postpone

PR #319, opened 2023-10-07
feat: add addMany function

LordMidas:
Allows adding multiple items to the weighted container with the same weight. Returns the container so it can be used as a pseudo-constructor.

### DIFF
--- a/msu/classes/weighted_container.nut
+++ b/msu/classes/weighted_container.nut
@@ -79,13 +79,34 @@
 		}
 	}
 
-	function addMany( _weight, _array )
+	// We expect 1 or 2 args.
+	// - When 1 it is an array
+	// - When 2 then the first arg is the weight and the second is the array
+	function addMany( ... )
 	{
-		::MSU.requireArray(_array);
-		foreach (item in _array)
+		// vargv[0] is "this"
+		if (vargv.len() == 1 || vargv.len() > 3)
 		{
-			this.add(item, _weight);
+			::logError("addMany accepts 1 or 2 arguments. If 1, then it is an array. If 2 then the first is the weight, and the second is the array");
+			throw "wrong number of parameters";
 		}
+
+		if (vargv.len() == 2)
+		{
+			foreach (item in vargv[1])
+			{
+				this.add(item);
+			}
+		}
+		else
+		{
+			local weight = vargv[1];
+			foreach (item in vargv[2])
+			{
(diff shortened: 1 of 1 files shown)


Enduriel (approved):

msu/classes/weighted_container.nut
```diff
@@ -79,6 +79,39 @@
 		}
 	}
 
+	// We expect 1 or 2 args.
+	// - When 1 it is an array
+	// - When 2 then the first arg is the weight and the second is the array
+	function addMany( ... )
```
TaroEld, line 85: Why does this use vargv and not just array as first parameter, and weight with a default value of 1.0 as the second parameter?

LordMidas: Because that is very bad for readability. When you have to pass a long array, and then a small weight at the end, it is very easy to miss that second parameter. Example:
```squirrel
local container = ::MSU.Class.WeightedContainer().addMany([
   something,
   something else,
   some morethings,
   even more stuff, 
   stuff,
   some other stuff,
   another stuff,
   some stuff,
   yes,
   more, 
   stuff],
   3);
```
vs
```squirrel
local container = ::MSU.Class.WeightedContainer().addMany(3, [
   something,
   something else,
   some morethings,
   even more stuff, 
   stuff,
   some other stuff,
   another stuff,
   some stuff,
   yes,
   more, 
   stuff]
```

TaroEld: For me it's quite readable both ways, you just need to know if you look at the beginning or the end. It also conceptually makes more sense to me to have the stuff to add come first, and the weight come later, so it helps usability. I also quite dislike common sense values you have to pass just because, such as a default 1 in this case. Plus, it makes the function less fugly.

LordMidas: You don't have to pass `1` in the case of vargv. 1 is the default if you just pass an array.

Enduriel: I don't really mind either way, (either weight first and require or weight second and optional) but I'm actually not a huge fan of the vargv approach either

Enduriel (changes requested):
As taro pointed out, I think either having weight first and be required or second and be optional is fine, but upon further reflection I don't like the vargv approach either.

TaroEld (approved):

TaroEld (approved):

PR #320, opened 2023-10-07
feat: allow making a weighted container ordered

LordMidas:
The primary purpose is to allow a `foreach (item in container)` type iteration and ` for` loop iteration with indices instead of the current `foreach (item, weight in container)`. This helps the container to be used interchangeably with arrays without having to do a `typeof` check and duplicate `foreach` loops separately for arrays and weighted containers.

I ran into this issue while coding the Dynamic Spawns Framework whereby the same member variable in a class could be an array or a weighted container. In many functions I want to iterate over that variable in `foreach` loops and sometimes in `for` loops with using `i` as index while being agnostic to its type. Having the possibility of an ordered weighted container allows me to set it as ordered during creation, and then the class can do its thing while not worrying about it.

While, in theory, we could implement a child class of WeightedContainer that is ordered (I did that in Dynamic Spawns and called it WeightedArray) - but that violates the Liskov Substitution Principle, which is why I'd like to have a basic implementation of this in the base class.

Alternatively, we implement a base abstract class and then inherit WeightedContainer from it and WeightedArray from it separately. Because ideally I'd like the WeightedArray to support all kinds of array functions e.g. `pop()`, `push()` etc.

### DIFF
--- a/msu/classes/weighted_container.nut
+++ b/msu/classes/weighted_container.nut
@@ -2,6 +2,7 @@
 {
 	Total = null;
 	Table = null;
+	Array = null;
 	Forced = null;
 	NextIItems = null;
 	NextIIndex = null;
@@ -16,7 +17,8 @@
 
 	function _get( _item )
 	{
-		if (_item in this.Table) return this.Table[_item];
+		if (this.Array != null) return this.Array[_item]; // _item is an idx in this case
+		else if (_item in this.Table) return this.Table[_item];
 		throw null;
 	}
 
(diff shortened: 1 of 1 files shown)


Enduriel:
We discussed and this is cool but as you said it should be a sibling of WeightedContainer and seems like it requires still quite a bit of work

Suor:
If you don't really care about order but just want to iterate over `index, item` instead of `index, weight` then you may just add an iterator method to it:

```squirrel
class WeightedContainer {
    ...    
    function items() {
        foreach (k, _ in this.Table) yield k;
    }
}

// Use
local items = type cont == "array" ? cont : cont.items();
foreach (item in items) {

}
```

Unfortunately typeof/items cannot be removed completely because `WeightedContainer` already has a way to iterate. Can pass `cont.items()` to functions waiting for arrays in many cases though.

PR #321, opened 2023-10-07
feat: add MSU.Class.Event class

LordMidas:
This emulates event listeners. Now someone can raise an event without having to know who is listening to the event.

### DIFF
--- a/msu/classes/event.nut
+++ b/msu/classes/event.nut
@@ -1,0 +1,72 @@
+::MSU.Class.Event <- class
+{
+	__Functions = null;
+
+	constructor()
+	{
+		this.__Functions = [];
+	}
+
+	function register( _env, _function )
+	{
+		// ::MSU.requireFunction(_function);
+		foreach (i, funcInfo in this.__Functions)
+		{
+			if (funcInfo[1] == _function)
+			{
+				return;
+			}
+		}
+
+		this.__Functions.push([_env, _function]);
+	}
+
+	function unregister( _function )
+	{
+		// ::MSU.requireFunction(_function);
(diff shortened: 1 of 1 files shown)


msu/classes/event.nut
```diff
@@ -0,0 +1,72 @@
...
+
+		foreach (funcInfo in this.__Functions)
+		{
+			_argsArray[0] = funcInfo[0];
+			funcInfo[1].acall(_argsArray);
+		}
+	}
+
+	// Will iterate over every registered function and check _conditionFunction
+	// if _conditionFunction returns false, stops iteration immediately
```
TaroEld, line 52: Feels to me like this should perhaps be based on the actual functions that are being called,ie if they return `false` then stop or something along those lines.

TaroEld:
Do we have an usecase for this?

msu/classes/event.nut
LordMidas: Nah, those functions can be anything. They don't necessarily have to be Boolean returning functions. This `conditionalInvoke` is more along the lines of replicating the `isAlive` feature of skill_container events where the iteration stops if the actor is no longer alive.

LordMidas:
> Do we have an usecase for this?

I mean event listeners are a basic thing in programming and I'm sure people can find some use case for this. It will help modders write more decoupled code.

msu/classes/event.nut
TaroEld: Returning false to stop an event from continuing feels quite natural, it's how it's done in JS event bubbling for example. Using some otherwise decoupled thing to check for a state is weird in contrast.

PR #322, opened 2023-10-20
feat: add `uniques` function to array utilities

LordMidas:


TaroEld (approved):

msu/utils/array.nut
```diff
@@ -104,4 +104,15 @@
 			}
 		}
 	}
+
+	// Returns a new array
+	function removeDuplicates( _array )
```
TaroEld, line 109: Actually this should probably be called `uniqueValues` or `asSet` or something, as it  does not change the array in place.

LordMidas: Yeah I see your point. `removeDuplicates` does sound like a function that should do its job in place. I don't like `asSet`. How about `filterUniques`?

TaroEld: `filterUniques` sounds like it'll only give you duplicate values, lol.

LordMidas: How about `uniques` or `getUniques`? I think `uniques` might be ok?

Enduriel: I think `.uniques()` is fine actually, same idea as `.keys()` and `.values()` which I like

msu/utils/array.nut
```diff
@@ -104,4 +104,15 @@
...
 		}
 	}
+
+	// Returns a new array
+	function uniques( _array )
+	{
+		local t = {};
+		foreach (a in _array)
+		{
+			t[a] <- true;
```
LordMidas, line 114: This will fail if an element in the array is `null`. So we need to add handling for that. I'll work on that next.

LordMidas: I have added handling for `null`. However it does not preserve the order in which `null` existed in the original array.

Enduriel (approved):

TaroEld:
Needs the soluttion we discussed

msu/utils/array.nut
LordMidas: I've pushed a new commit based on the algorithm we decided on that preserves order and handles nulls.

TaroEld (approved):

PR #316, opened 2023-09-22
Feat port to modern hooks

LordMidas:

PR #323, opened 2023-10-20
feat: add getItemsByFunction and getItemsByFunctionAtSlot

LordMidas:
To be merged AFTER #316 (port to modern hooks)

### DIFF
--- a/msu/hooks/items/item_container.nut
+++ b/msu/hooks/items/item_container.nut
@@ -96,29 +96,15 @@
 	q.getItemsByFunction <- function( _function )
 	{
 		local ret = [];
-
 		foreach (slot, _ in ::Const.ItemSlotSpaces)
 		{
-			this.__putItemsAtSlotIntoArray(slot, ret);
+			ret.extend(this.getItemsByFunctionAtSlot(slot, _function));
 		}
-
 		return ret;
 	}
 
 	q.getItemsByFunctionAtSlot <- function( _slot, _function )
 	{
-		local ret = [];
-		this.__putItemsAtSlotIntoArray(_slot, ret);
-		return ret;
-	}
-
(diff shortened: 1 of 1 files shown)


msu/hooks/items/item_container.nut
```diff
@@ -94,4 +94,35 @@
...
 		return unequip(_item);
 	}
+
+	q.getItemsByFunction <- function( _function )
+	{
+		local ret = [];
+
+		foreach (slot, spaces in ::Const.ItemSlotSpaces)
+		{
+			for (local i = 0; i < spaces; i++)
```
TaroEld, line 104: This should probably just use `getItemsByFunctionAtSlot`:
`ret.extend(getItemsByFunctionAtSlot(slot, _function);`

LordMidas: That results in way more iterations and array declarations than the current implementation. I don't like that.

TaroEld: Is it worse than resusing the same code in two functions with the only difference being another loop though

TaroEld: nvm

msu/hooks/items/item_container.nut
```diff
@@ -92,4 +92,36 @@
...
+			{
+				this.__putItemsAtSlotIntoArray(slot, ret);
+			}
```
TaroEld, line 105: Wait, I don't get it. Why this inner loop? Shouldn't this just run for all the slots, and the private function then iterates over the items in that slot?

LordMidas: Fixed

TaroEld (approved):

TaroEld (approved):

PR #331, opened 2023-12-06
feat: add persistent storage which doesn't require BBParser

Enduriel:
So this is a big one.

Basic interface is of course in the persistent_data_mod_addon, should be self-explanatory.

The backend stuff is a bit more interesting. Basically, I emulate the entire backend system serialization system so that we can have a soft error in case of issues and the game doesn't hard crash if a mod messes something up. The performance of this is frankly still pretty great and not an issue at all in my testing (though I admittedly haven't quantified how long stuff takes)

The way the data is structured is a bit strange but here's the basic concept.
Data is serialized by storing the type followed by the data, so when deserializing we can always safely decode it and identify when we have a type we don't know. When deserializing we immediately instantiate the correct serialiation data type for the type we read immediately after reading the type.

More complex data inherits from `CustomSerializationData `whereas basic data inherits from `BasicSerializationData`. `CustomSerializationData `has the main chunk of the (de)serialization logic, because each `SerializationData `is responsible for it's own (de)serialization, and since `CustomSerializationData `must be able to (de)serialize all data, it must contain the logic for that.

At the top level, when createFile is called with some data, it is used to instantiate an `ArraySerializationData` which then builds its internal array with all the passed data converted into `SerializationData` and prepared for serialization. The one exception is when (de)serializing BB Objects, those must use a `DataArrayData` to store and read their data. Imo this is best explained with a simple example:

```squirrel
local myData = {
    string = "someString",
    int = 124,
    float = "1.1241241",
    player = ::MSU.Class.DataArrayData(),
    table = {
        nestedInt = 1
    }
}

local myPlayer //create or grab player
myPlayer.onSerialize(myData.player.getSerializationEmulator()) // fill the DataArrayData with player data

mod.PersistentData.createFile("someName", myData) // the created file will be MSU#mod_id#someName
// it will exist in game saves but will not be shown in load/save screens due to the hook on UIDataHelper
// the data will first be fully converted using the StrictSerializationEmulator
// to ensure there will be no hard errors during file creation
// and then that will write it

// now to get data back we simply do
local readData = mod.PersistenData.readFile("someName")
// this should now equal to myData
// to get player back we can do
local readPlayer // create player from temp roster or smtng
readPlayer.onDeserialize(readData.player.getDeserializationEmulator())
```
The point is that except for DataArrayData, all the data can be infinitely nested safely without issues, which allows for completely arbitrary data organization.

The StrictSerDeEmulators also have some pretty cool uses for debugging stuff like the legends save corruption because they allow you to basically serialize the entire game into a single DataArrayData safely, guarantee either a useful error or a safe save to an external file which could then be read in a different instance which imo has huge implications.

Necro has also obviously been testing this in his origin maker which (along with MSU's mod settings) are the two major mods which would immediately hugely benefit from this system.

### DIFF
--- a/msu/systems/serialization/serialization_types/custom_serialization_data.nut
+++ b/msu/systems/serialization/serialization_types/custom_serialization_data.nut
@@ -19,7 +19,7 @@
 	// must be called when overriden
 	function serialize( _out )
 	{
-		_out.writeU8(this.getType());
+		base.serialize(_out);
 		::MSU.Class.U32SerializationData(this.len()).serialize(_out); // store length
 	}
 
--- a/msu/systems/serialization/serialization_types/custom_serialization_data.nut
+++ b/msu/systems/serialization/serialization_types/custom_serialization_data.nut
@@ -71,7 +71,7 @@
 				array.deserialize(_in);
 				return array;
 			default:
-				local unknownData =  ::MSU.Class.UnknownSerializationData(type, null);
+				local unknownData =  ::MSU.Class.UnknownSerializationData(type);
 				unknownData.deserialize(_in);
 				return unknownData;
 		}
@@ -125,7 +125,7 @@
 			case "table":
 				if (::MSU.isBBObject(_value))
 				{
-					::logError("MSU Serialization cannot handle BB Objects");
+					::logError("MSU Serialization cannot serialize BB Objects directly");
 					throw ::MSU.Exception.InvalidValue(_value);
 				}
 				return ::MSU.Class.TableSerializationData(_value);
--- a/msu/systems/serialization/serialization_types/custom_serialization_data.nut
+++ b/msu/systems/serialization/serialization_types/custom_serialization_data.nut
@@ -1,0 +1,45 @@
+::MSU.Class.CustomSerializationData <- class extends ::MSU.Class.AbstractSerializationData
+{
+	__MetaData = null;
+
+	constructor(_data, _metaData)
+	{
+		this.__MetaData = _metaData;
+		base.constructor(_data);
+	}
+
+	function getMetaData()
+	{
+		return this.__MetaData;
+	}
+
+	function setMetaData( _metaData )
+	{
+		this.__MetaData = _metaData;
+	}
+
+	function deserialize( _in )
+	{
+		local type = _in.readU8();
+		if (type != this.DataType.U32)
+			throw ::MSU.Exception.InvalidValue(type);
+		this.setLength(_in.readU32());
+	}
+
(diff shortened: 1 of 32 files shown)


msu/systems/serialization/strict_deserialization_emulator.nut
```diff
@@ -0,0 +1,66 @@
+// emulates the _in object passed to onDeserialize functions
+::MSU.Class.StrictDeserializationEmulator <- class extends ::MSU.Class.StrictSerDeEmulator
```
LordMidas, line 2: I could not understand fully the need for these new ser/de emulator classes and would like to find a way to use the original ones we already have. But my brain is rather slow atm and it seems this isn't the main point of the PR so I'll talk to you on VC about this - don't want to delay my review for this.

msu/systems/serialization/serialization_types/data_array_data.nut
```diff
@@ -0,0 +1,60 @@
+::include("msu/systems/serialization/serialization_types/raw_data_array_data");
+::MSU.Class.DataArrayData <- class extends ::MSU.Class.RawDataArrayData
```
LordMidas, line 2: What's the point of this class?

LordMidas (changes requested):

msu/systems/serialization/serialization_types/custom_serialization_data.nut
```diff
@@ -0,0 +1,147 @@
...
+			case this.DataType.I32:
+				return ::MSU.Class.I32SerializationData(_in.readI32());
+			case this.DataType.F32:
+				return ::MSU.Class.F32SerializationData(_in.readF32());
+			case this.DataType.DataArray:
+				local dataArray = ::MSU.Class.DataArrayData();
+				dataArray.deserialize(_in);
+				return dataArray;
+			case this.DataType.Table:
+				local table = ::MSU.Class.TableSerializationData(null);
```
LordMidas, line 66: Ah here you pass null so that's why you need `null` handling in the constructors. Instead of that I'd just pass an empty table or array here as needed and remove the null handling from the constructor in those classes.

msu/systems/serialization/serialization_types/data_array_data.nut
LordMidas: Based on our VC it became clear to me that the point of this class is to have an intermediary data representation that is distinct from array/table in the intermediary state. So yes, we need the class, but we should rename it to something else.

Enduriel:
Superseded by #341

PR #332, opened 2024-01-05
Taro storage persistence

TaroEld:
Ignore the commit names.
Main changes:
- Remove the individual primitives and replace them with a single one that gets passed its type (actually not 100% on this one)
- Move some things to the system instead of the classes (data type enum, read/write function defs, initializing values, validation)
- Merge BasicSerializationData and AbstractSerialisationData to PrimitiveSerializationData
- rename CustomSerializationData to SerializationDataCollection; also no longer inherits from abstract. Very slight code duplication, but I think it's worth it.
- simplify SerializationDataCollection and children


msu/systems/serialization/serialization_types/primitive_serialization_data.nut
```diff
@@ -0,0 +1,41 @@
...
+
+	function getWriteFunctionString()
+	{
+		return ::MSU.System.Serialization.WriterFunctionStrings[this.getType()];
+	}
+
+	function getReadFunctionString()
+	{
+		return ::MSU.System.Serialization.ReaderFunctionStrings[this.getType()];
+	}
```
TaroEld, line 40: Upon returning I have no clue why I put these in the system and not just here

msu/systems/serialization/debug_serde_emulator.nut
```diff
@@ -0,0 +1,50 @@
+// enulates the _out object passed to onSerialize functions
+::MSU.Class.DebugSerdeEmulator <- class
+{
```
TaroEld, line 3: These we can still toss out, was more for testing the implementation

msu/systems/serialization/serialization_types/primitive_serialization_data.nut
TaroEld: Okay it was because the system needs "ReaderFunctionStrings"

Enduriel:
superseded by #341

PR #337, opened 2024-02-26
feat: add SQ keybind input blocker when a JS input or textarea field is focused

TaroEld:


Enduriel:
Some thoughts,

1. Does this actually benefit from being a setting? Would it be better to just always have this on? Are there times when we wouldn't want this to be enabled?
2. If it does remain a setting, we might want to give the setting a more descriptive ID and have modders handle it as a setting JS side (ie disable it directly with `setModSetting`), and also internally don't add a separate msu connection function and just use the setting directly
3. Finally bit of an edge case consideration but what happens if the setting is changed while someone is typing right now, pretty sure it affect anything until someone exits and re-enters the text field, is that desirable?

TaroEld:
The setting is a 'disable' failsafe, should this break some mod in some way I hadn't considered. It might not be necessary, but I'd rather have it and not need it. It's for users, not modders.It'll enable SQ keybinds if changed, I think that suffices.On Feb 27, 2024 10:43, Enduriel ***@***.***> wrote:
Some thoughts,
Does this actually benefit from being a setting? Would it be better to just always have this on? Are there times when we wouldn't want this to be enabled?If it does remain a setting, we might want to give the setting a more descriptive ID and have modders handle it as a setting JS side (ie disable it directly with setModSetting), and also internally don't add a separate msu connection function and just use the setting directlyFinally bit of an edge case consideration but what happens if the setting is changed while someone is typing right now, pretty sure it affect anything until someone exits and re-enters the text field, is that desirable?

TaroEld:
This branch now also includes changes to enable continuous keybinds firing regularly, re-adds camera keybinds, and fixes an issue with a tactical state keybind.

TaroEld:
This branch no longer includes the continuous keybind changes.

msu/systems/keybinds/keybind_sq.nut
```diff
@@ -16,6 +17,12 @@
 		this.KeyState = _keyState;
 	}
 
+	function setBypassInputDenied(_bool)
+	{
+		this.BypassInputDenied = _bool;
+		return this;
```
Enduriel, line 23: As much as I agree with returning this here in principle, we haven't been doing this. I think this would be a useful thing to refactor in the codebase as a whole but doing this in just this one function seems wrong to me, so it should be a separate PR/commmit

Enduriel (changes requested):

Enduriel (approved):

LordMidas (approved):
I don't know anything about this so you and Enduriel go ahead.

PR #338, opened 2024-02-27
refactor: change certain functions to only when alive

LordMidas:
Some mods may kill certain entities during these functions, e.g. during onMovementFinished because of a certain skill/effect on the entity, which may then cause subsequently updated skills to crash. This change prevents such cases.

Example of a problem that is intended to be fixed by this PR: https://discord.com/channels/855921501286039582/855921501286039585/1210475458165084241

### DIFF
--- a/msu/hooks/skills/skill_container.nut
+++ b/msu/hooks/skills/skill_container.nut
@@ -89,7 +89,7 @@
 
 	q.onMovementFinished <- function( _tile )
 	{
-		this.callSkillsFunction("onMovementFinished", [
+		this.callSkillsFunctionWhenAlive("onMovementFinished", [
 			_tile
 		]);
 	}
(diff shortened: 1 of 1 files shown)


TaroEld (approved):

Enduriel (approved):
If this is fixing crashes without affecting 'working' code, then this should be correctly marked as a fix. If that is changed lgtm

LordMidas:
Actually on further deliberation I am not sure if this needs to be done. Looking at vanilla skill event functions, there are very few which have the `isAlive` check. I think ideally the modder should not kill the character during these functions.

In any case, there's also the issue of for example if a skill does the following
```squirrel
function onMovementFinished( _tile )
{
    ::Const.SomeGlobalCounter++;
}
```
If we make this event alive only then this function won't be called for this skill if a skill before this skill kills the target during that skill's `onMovementFinished`.

So it isn't immediately clear what the best course of action here is. So I'd wait on this PR for now.

PR #339, opened 2024-02-28
feat: add deepEquals function

Enduriel:
Inspired by Hackflow's std lib deepEq which does the same thing. I ended up needing this for some assertions when unit testing modern hooks

### DIFF
--- a/msu/utils/utils.nut
+++ b/msu/utils/utils.nut
@@ -38,6 +38,43 @@
 		}
 	}
 
+	function deepEquals(_a, _b)
+	{
+		if (_a == _b)
+			return true;
+		if (typeof _a != typeof _b)
+			return false;
+		if (typeof _a != "array" && typeof _a != "table")
+		{
+			return false;
+		}
+		if (_a.len() != _b.len())
+		{
+			return false;
+		}
+		if (typeof _a == "array")
+		{
+			foreach (i, v in _a)
+			{
+				if (!this.deepEq(v, _b[i]))
+				{
+					return false;
+				}
+			}
+		}
+		else
+		{
+			foreach (k, v in _a)
+			{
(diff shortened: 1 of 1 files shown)


msu/utils/utils.nut
```diff
@@ -38,6 +38,35 @@
 		}
 	}
 
+	function deepEquals(_a, _b)
```
LordMidas, line 41: Space before/after params.

msu/utils/utils.nut
LordMidas, line 1: 1. I think this should exist in `globals` as `::MSU.isDeepEqual` in the same vein as `::MSU.isEqual`. Alternatively add a 3rd parameter to `::MSU.isEqual( _a, _b, _deep = false )` which is false by default. And move this functionality to that function.
2. It should handle instances and classes.
3. It should handle WeakTableRef.

I would rewrite the entire function body as:
```squirrel
switch (typeof _a)
{
	case "array":
		if (typeof _b == "array")
		{
			foreach (i, v in _a)
			{
				if (!this.deepEq(v, _b[i]))
					return false;
			}
		}
		return true;

	case "table":
		if (typeof _b == "table")
		{
			foreach (k, v in _a)
			{
				if (!(k in _b) || !this.deepEq(k, _b[k]))
					return false;
			}
		}
		return true;

	case "instance":
		if (_a instanceof ::WeakTableRef && _b instanceof ::WeakTableRef)
			return this.deepEq(_a.get(), b.get());

		if (typeof _b != "instance" || !(_b instanceof _a.getclass()))
			return false;

		foreach (k, v in _a.getclass())
		{
			if (typeof v == "function")
				continue;

			if (a[k] != b[k])
				return false;
		}
		return true;

	default:
		return _a == _b;
}
```

LordMidas (changes requested):

PR #341, opened 2024-03-18
Storage persistence midas 2

LordMidas:
- Cleaned up classes. Now we have a base `AbstractData` class which has 2 children:
   - `PrimitiveData` which is for value types like integer, float etc.
   - `ArrayData`. This further has 2 children:
       - `TableData`
       - `SerializationData` which is used in emulators to hold the ser/de data in the form of instances of these above classes.

- Removed most classes including `RawDataArrayData` and `DataArrayData`. Their functionality is now handled directly via `SerializationData` which is also now embedded into `SerDeEmulator` as a data container.

- Removed `StrictSerDe` emulators and instead add type-checking directly to the regular ones.

- Deprecated `Utils.serialize`, `Utils.deserialize` and `Utils.deserializeInto` functions and added a new `::MSU.Serialization` namespace which now contains improved alternatives for these functions.

- Moved the serialization data types enum to the new `::MSU.Serialization` namespace.

- Added `DeserializeMetaData` key to `SerializationSystem` which is set during `onBeforeDeserialize` of `world_state`. We then use this metadata during `flagDeserialize` of the serialization system so that `isSavedVersionAtLeast` etc. report the correct version with this metadata. I'd consider this a bugfix in MSU unless there is something wrong with this approach.

- Changed inheritance structure of flag ser/de emulators from the base `SerDeEmulator` class. This also renames the previously known `SerializationEmulator` and `DeserializationEmulator` into the more appropriate `FlagSerializationEmulator` and `FlagDeserializationEmulator`. This also involves moving the `read/write` funtions directly into the parent class and allowing the child classes to instead modify the relevant `__readData` and `__writeData` functions and add other funtionality as desired. This was important because I wanted a `SerializationEmulator` and `DeserializationEmulator` class which can be instantiated from the `SerializationData` without having to pass in a Mod Id and Flag Container. So instead one can simply do `::MSU.Class.SerializationEmulator(::MSU.Class.MetaDataEmulator(), this)` inside the `SerializationData`.

The last one above is probably the most controversial change in this PR. I'd like to discuss it and am open to feedback / suggestions for alternative approaches.

Note that because of this change to the emulators, after updating to MSU 1.4.0 when people load their older saved game for the first time, the log will be spammed with warnings of mismatch between read data and the data type stored in the emulator when deserializaing mod settings. However, the values loaded will still be correct. This won't happen for saves made with MSU 1.3.0.

Fully tested and working with the following example code:
```squirrel
local mod = ::MSU.Mod;

local myData = {
    string = "someString",
    int = 124,
    float = "1.1241241",
    player = ::MSU.Class.SerializationData()
    table = {
        nestedInt = 1
    }
}

::getBro("Ruthard").onSerialize(myData.player.getSerializationEmulator());

mod.PersistentData.createFile("testPersistence", myData);

local readData = mod.PersistentData.readFile("testPersistence");
local readPlayer = ::World.getTemporaryRoster().create("scripts/entity/tactical/player");
readPlayer.onDeserialize(readData.player.getDeserializationEmulator());
::logInfo(readPlayer.getName());
```


msu/systems/serialization/flag_serde_emulator.nut
```diff
@@ -0,0 +1,37 @@
...
+
+	constructor(_mod, _id, _flagContainer, _metaDataEmulator = null)
+	{
+		if (_metaDataEmulator == null) _metaDataEmulator = clone ::MSU.System.Serialization.MetaData;
+		if (this.__IDRegex.match(_id))
+		{
+			::logError("the ID passed to flag serialization cannot end with a full stop followed by digits so it doesn't collide with internal MSU flags");
+			throw ::MSU.Exception.InvalidValue(_id);
+		}
+		base.constructor(_metaDataEmulator)
```
Enduriel, line 17: missing `;` I haven't been checking every line but if there's one there's probably more

msu/systems/serialization/serde_emulator.nut
```diff
@@ -1,45 +1,170 @@
...
 ::MSU.Class.SerDeEmulator <- class
 {
-	static __IDRegex = regexp("^.*\\.\\d+$")
-	Data = null;
-	Mod = null;
-	ID = null;
 	MetaData = null;
-	FlagContainer = null;
+	SerializationData = null;
+	Idx = -1;
```
Enduriel, line 6: This should be moved into the deserialization emulators only, so into `__ReadFunctions`, which means that it should probably be renamed into `__ReadFields`

msu/systems/serialization/serde_emulator.nut
```diff
@@ -1,45 +1,170 @@
...
+			return null;
+		}
+		local data = this.SerializationData.getDataArray()[this.Idx];
+		if (data.getType() != _type)
+		{
+			::logError(format("The type being read %s isn't the same as the type %s stored in the Deserialization Emulator", ::MSU.Serialization.DataType.getKeyForValue(_type), ::MSU.Serialization.DataType.getKeyForValue(this.SerializationData.getDataArray()[this.Idx].getType())));
+			// currently still continuing in case of conversion between integers
+		}
+		return data.getData();
+	}
```
Enduriel, line 169: Also these should be in `__XXXXFunctions` unless I'm missing something

msu/systems/serialization/serialization_types/abstract_data.nut
```diff
@@ -0,0 +1,31 @@
...
+	{
+		return this.__Data;
+	}
+
+	function serialize( _out )
+	{
+		_out.writeU8(this.__Type);
+	}
+
+	// Must be overridden by children
```
Enduriel, line 27: should this throw?

msu/utils/serialization.nut
```diff
@@ -0,0 +1,132 @@
...
+		"U16",
+		"U32",
+		"I8",
+		"I16",
+		"I32",
+		"F32",
+		"Table",
+		"Array",
+		"SerializationData"
+	]),
```
Enduriel, line 18: This is still currently a problem, should maybe comment here that this should not be extended. This will currently break if extended by multiple mods because there is no ordering guarantee, the solution is to serialize and deserialize this enum alongside any data (probably into metadata). I assume we won't be doing this so instead comment here and in doc that this shouldn't be extended

msu/utils/utils.nut
```diff
@@ -13,31 +13,7 @@
...
-			case "table":
-				return this.DataType.Table;
-			case "class":
-				return this.DataType.Class;
-			case "instance":
-				return this.DataType.Instance;
-			case "null":
-				return this.DataType.Null;
-		}
-	}
```
Enduriel, line 39: Surprised we can get away with this but it does seem like we can

Enduriel (changes requested):

msu/systems/serialization/serialization_types/abstract_data.nut
LordMidas: No one should be instantiating this class. We could add a throw just to make it absolutely clear LOL but idk if that is necessary.

msu/systems/serialization/serde_emulator.nut
LordMidas: Children have to overwrite these functions so I couldn't find a better way to do it. Open to suggestions.

msu/utils/serialization.nut
LordMidas: Based on our discussion we decided to postpone adding the ability to extend this enum to a future MSU release. For now, we will document it that modders must not extend it.

msu/systems/serialization/deserialization_emulator.nut
```diff
@@ -1,88 +1,9 @@
...
-			if (!this.FlagContainer.has(currentFlag))
-				return false;
-			this.Data[i] = this.FlagContainer.get(currentFlag);
-			this.FlagContainer.remove(currentFlag);
-		}
-		return true;
-	}
+foreach (key, value in ::MSU.Class.SerDeEmulator.__ReadFunctions)
+{
+	::MSU.Class.DeserializationEmulator[key] <- value;
```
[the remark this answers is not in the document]
LordMidas: I understand but don't think the alternative is better. Putting it in another table creates spaghetti and one then has to ensure that file/table gets loaded before these files. There's also the issue of which namespace to put that in etc. I just think this is simple and the comments are explaining the purpose of these tables.

msu/systems/persistent_data/persistent_data_system.nut
```diff
@@ -108,4 +109,46 @@
 		result += this.Separator
 		::logInfo(result);
 	}
+
+	function createFile( _fileName, _dataArray )
+	{
+		::MSU.requireInstanceOf(::MSU.Class.ArrayData, _dataArray);
```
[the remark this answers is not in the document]
LordMidas: It's an internal function though because it is called from the system addon. However, if we want this function to be generally callable by other than that one caller from the addon then yes we can change it. Shall we?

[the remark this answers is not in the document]
Enduriel: I don't think we need to change it, but this is just the wrong type here, either remove the require or make it abstract.

Enduriel:
You typoed %s and instead wrote $s in your commit moving MSU# into a separate table

msu/systems/persistent_data/persistent_data_system.nut
```diff
@@ -108,4 +110,46 @@
...
+	}
+
+	function validateFileName( _fileName )
+	{
+		if (!this.FileNameRegexp.match(_fileName))
+		{
+			::logError("Battle Brothers file saves can only contain the characters 'a-zA-Z0-9_,-+# '");
+			throw ::MSU.Exception.InvalidValue(_fileName);
+		}
+	}
```
TaroEld, line 131: Seems to be only used in that one spot and is not very long, so just move it into the function?

msu/systems/serialization/serialization_types/array_serialization_data.nut
```diff
@@ -0,0 +1,55 @@
+::MSU.Class.ArrayData <- class extends ::MSU.Class.AbstractData
+{
+	__DataArray = null;
+
+	constructor( _data )
```
TaroEld, line 5: In your example you pass an empty array here - just have a default null and initialize it here?

TaroEld (approved):
I assume you tested thoroughly and also tested that flag containers still work and survive the conversion between versions.

msu/systems/serialization/serialization_types/array_serialization_data.nut
LordMidas: Sorry the example was not updated. It should be instantiating `::MSU.Class.SerializationData()` and that can definitely be initialized without passing any argument as it defaults to null and instantiates the array if null. For the `ArrayData` and `TableData` I require passing an array/table because they are meant to convert the given array/table to the data and do not have the possibility to later add data to it. The `SerializationData` class is the one that can take data after it has been instantiated.

msu/systems/persistent_data/persistent_data_system.nut
LordMidas: @Enduriel

msu/systems/persistent_data/persistent_data_system.nut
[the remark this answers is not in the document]
LordMidas: To clarify, by "change it" I meant make the change you are suggesting i.e. require instance of AbstractData. My point was that because we are using it only internally we just check for the type we send from the addon function. However, if you think this function can be called from places other than the addon, then it makes sense to require instance of AbstractData.

msu/systems/persistent_data/persistent_data_system.nut
Enduriel: @TaroEld do you mean the `FileNameRegexp`? I don't think we should be recreating it every function call if we can cache it

Enduriel (approved):
As Taro said, this is assuming everything is tested and save compatible

msu/systems/persistent_data/persistent_data_system.nut
TaroEld: I meant the whole validation function, but it doesn't really matter.

Enduriel (approved):
Looks reasonable to me, if this is what you have to do to get it working with vanilla then it's kinda out of our hands

Enduriel:
After some discussion in Discord we still have to do the following:

- rework MetaDataEmulator to be constructed with a table/instance of dummy/real metadata and serialize all relevant information from itself and that real metadata @Enduriel 
- rework the SerializationData to serialize the new MetaDataEmulator so that persistently serialized bb classes will automatigically have the right metadata attached to them @LordMidas 
- throwing vs erroring during deserialization type mismatch -> error and print row & file like in printData @LordMidas

msu/hooks/states/world_state.nut
```diff
@@ -190,6 +190,7 @@
 	q.onBeforeSerialize = @(__original) function( _out )
 	{
 		__original(_out);
+
```
LordMidas, line 193: This extra line can be removed from this PR to keep its diff cleaner.

msu/hooks/states/world_state.nut
```diff
@@ -331,3 +332,27 @@
 		return __original(_mouse);
 	}
 });
+
+::MSU.QueueBucket.VeryLate.push(function() {
+	::MSU.MH.hook("scripts/states/world_state", function(q) {
+		q.onBeforeSerialize = @(__original) function( _out )
+		{
+			::MSU.System.Serialization.SerializationMetaData = _out.getMetaData();
+			::MSU.System.Serialization.MidOnBeforeSerialize = true;
```
LordMidas, line 341: I prefer a name like `IsDuringOnBeforeSerialize`.

msu/hooks/states/world_state.nut
```diff
@@ -331,3 +332,27 @@
...
+		q.onBeforeDeserialize = @(__original) function( _in )
+		{
+			::MSU.System.Serialization.DeserializationMetaData = _in.getMetaData();
+			__original(_in);
+		}
+
+		q.onDeserialize = @(__original) function( _out )
+		{
+			__original(_out);
+			::MSU.System.Serialization.DeserializationMetaData = null;
```
LordMidas, line 355: I once again ask for sanity check if we have ensured that this `onDeserialize` is the last function to actually run during deserialization and, therefore, it is safe to set this back to null here?

msu/hooks/ui/global/data_helper.nut
```diff
@@ -0,0 +1,14 @@
+::MSU.MH.hook("scripts/ui/global/data_helper", function(q) {
+	q.convertCampaignStoragesToUIData = @( __original ) function()
+	{
+		local queryStorages = ::PersistenceManager.queryStorages;
+		::PersistenceManager.queryStorages = function()
+		{
+			return queryStorages().filter(@(_i, _v) !::MSU.String.startsWith(_v.getFileName(), ::MSU.System.PersistentData.FilePrefix));
```
LordMidas, line 7: `_i` should be `_`.

msu/systems/persistent_data/persistent_data_mod_addon.nut
```diff
@@ -14,4 +15,36 @@
 	{
 		::MSU.System.PersistentData.writeToLog(_fileID, this.Mod.getID(), _payload);
 	}
+
+	// MSU 1.3.0
+	function prefixFileName( _fileName )
```
LordMidas, line 20: This is a private function, right? So should be prefixed with `__`.

msu/systems/persistent_data/persistent_data_mod_addon.nut
```diff
@@ -14,4 +15,36 @@
...
+
+	function hasFile( _fileName )
+	{
+		return ::MSU.System.PersistentData.hasFile(this.prefixFileName(_fileName));
+	}
+
+	function getFiles()
+	{
+		local prefix = ::MSU.System.PersistentData.FilePrefix + this.Mod.getID();
+		return ::PersistenceManager.queryStorages().filter(@(_i, _v) ::MSU.String.startsWith(_v.getFileName(), prefix))
```
LordMidas, line 38: `_i` should be `_`.

msu/systems/persistent_data/persistent_data_system.nut
```diff
@@ -1,8 +1,10 @@
 ::MSU.Class.PersistentDataSystem <- class extends ::MSU.Class.System
 {
+	static FilePrefix = "MSU#";
 	Mods = null;
 	ModConfigPath = "mod_config/";
 	Separator = "@"
+	FileNameRegexp = regexp("^[\\w\\d\\,\\-\\+\\# ]+$")
```
LordMidas, line 7: missing `;`

msu/systems/serialization/metadata_emulator.nut
```diff
@@ -3,15 +3,23 @@
 {
 	Version = null;
 	Data = null;
+	RealMeta = null;
```
LordMidas, line 6: Should be renamed to `RealMetaData`

msu/systems/serialization/metadata_emulator.nut
```diff
@@ -3,15 +3,23 @@
...
+
 	constructor()
 	{
 		this.Data = {};
 		this.Version = ::Const.Serialization.Version;
 	}
 
-	function getVersion()
+	// this would ideally be an alternative static constructor
+	function setRealMeta( __realMeta )
```
LordMidas, line 19: `setRealMetaData`

LordMidas, line 19: Why is the parameter prefixed with `__`? We don't have a convention for this afaik.

msu/systems/serialization/serialization_system.nut
```diff
@@ -42,24 +44,28 @@
 		}
 		if (_flags == null) _flags = ::World.Flags;
 
-		local inEmulator = ::MSU.Class.DeserializationEmulator(_mod, _id, _flags);
+		local inEmulator = ::MSU.Class.FlagDeserializationEmulator(_mod, _id, _flags, ::MSU.System.Serialization.DeserializationMetaData);
```
LordMidas, line 47: The manual passing of the `DeserializationMetaData` is now unnecessary as we have added handling for null argument in the constructor of the `FlagDeserializationEmulator`.

msu/systems/serialization/serialization_types/primitive_data.nut
```diff
@@ -0,0 +1,85 @@
...
+			this.__Data = _in["read" + ::MSU.Serialization.DataType.getKeyForValue(this.getType())]();
+	}
+
+	function __printInvalidDataError( _type, _data )
+	{
+		::logError(format("Storing invalid or unexpected data \'%s\' in container of type: %s", _data + "", ::MSU.Serialization.DataType.getKeyForValue(_type)));
+		// We have to print the full stack trace here because we cannot know for sure which level of stackinfos will be the actual source of the problem
+		// as it will be different if this is being instantiated by someone in their own function somewhere or if it is being instantiated
+		// by MSU functions e.g. ::MSU.Serialization.__convertValueFromBaseType
+		::MSU.Log.printStackTrace();
```
LordMidas, line 83: @Enduriel , @TaroEld Please confirm if my reasoning and implementation is correct here.

msu/utils/serialization.nut
```diff
@@ -0,0 +1,160 @@
...
+				::logError("MSU Serialization cannot handle instances other than descendants of ::MSU.Serialization.Class.AbstractData");
+				throw ::MSU.Exception.InvalidValue(_value);
+
+			default:
+				::logError("Attempted to serialize unknown type");
+				throw ::MSU.Exception.InvalidType(_value);
+		}
+	}
+
+	function __convertValueFromGivenType( _value, _type )
```
LordMidas, line 137: Needs handling for Null data type

msu/systems/serialization/metadata_emulator.nut
```diff
@@ -40,56 +50,186 @@
...
+
+	function serialize( _out )
+	{
+		if (::MSU.System.Serialization.MidOnBeforeSerialize)
+		{
+			throw "Can't serialize metadata during onBeforeSerialize!";
+		}
+		local realMeta = this.RealMeta;
+		local knownValues = {};
+		local dummyMeta = {
```
LordMidas, line 159: @Enduriel I remember we added some comments here explaining the reasoning for overwriting functions. The comments are present in the commit on your other branch, but apparently wasn't copied over to the commit on this branch when you cherrypicked/pushed.

msu/systems/serialization/metadata_emulator.nut
```diff
@@ -40,56 +50,186 @@
...
+			};
+		}
+		return cleanup;
+	}
+
+	function serialize( _out )
+	{
+		if (::MSU.System.Serialization.MidOnBeforeSerialize)
+		{
+			throw "Can't serialize metadata during onBeforeSerialize!";
```
LordMidas, line 155: Better wording would be: `attempting to serialize metadata during world_state.onBeforeSerialize` to go in line with other default squirrel exceptions.

msu/systems/serialization/metadata_emulator.nut
```diff
@@ -40,56 +50,186 @@
...
+			cleanup = function() {
+				::World.Assets = null;
+				::Stash = null;
+				assets = null;
+			};
+		}
+		return cleanup;
+	}
+
+	function serialize( _out )
```
LordMidas, line 151: So this serialize function right now will always trigger the world_state onBeforeSerialize function and therefore copy all the values that are set in that function into this metadata. What if someone has a metadata emulator which he has populated with certain values and wants to now serialize that only? Perhaps we should have a separate function to handle these two separate cases?

msu/systems/serialization/metadata_emulator.nut
Enduriel: I typoed it apparently

msu/systems/serialization/load.nut
```diff
@@ -27,5 +30,6 @@ includeFile("serialization_mod_addon.nut");
 		_get = @(_k)@(...)""
 	});
 	worldState.onBeforeSerialize(serEm);
+	::MSU.System.Serialization.SerializationMetaData = serEm.getMetaData();
```
Enduriel, line 33: I'm not having issues without this line so I believe it is redundant due to the hook on onBeforeSerialize in world_state

msu/systems/serialization/metadata_emulator.nut
Enduriel: While I want to call that bloat, that does actually make some sense, but it has a knock-on affect with SerializationData, I think something like this would be better as a flag on SerializationData that then applies to the MetaData, so have a field rather than a separate function, this would also have to be serialized.

msu/systems/serialization/metadata_emulator.nut
Enduriel: Yeah, probably worth adding them back in

msu/systems/serialization/serialization_system.nut
```diff
@@ -2,7 +2,9 @@
 {
 	Mods = null;
 	EmulatorsToClear = null;
-	MetaData = null;
+	MidOnBeforeSerialize = false;
```
Enduriel, line 5: As you mentioned above, this should be renamed

msu/systems/serialization/serialization_types/array_data.nut
```diff
@@ -0,0 +1,55 @@
...
+
+	function len()
+	{
+		return this.__Data.len();
+	}
+
+	function getDataArray()
+	{
+		return this.__DataArray;
+	}
```
Enduriel, line 26: Why does this need to be public?

msu/systems/serialization/serialization_types/primitive_data.nut
Enduriel: Unfortunately that does seem to be right, and while the log spam is annoying in this case I can't think of an alternative either.

msu/systems/serialization/serialization_types/serialization_data.nut
```diff
@@ -0,0 +1,60 @@
...
+
+	function write( _element, _type )
+	{
+		this.__Data.push(_element);
+		this.__DataArray.push(::MSU.Serialization.__convertValueFromGivenType(_element, _type));
+	}
+
+	function serialize( _out )
+	{
+		this.__MetaData.setRealMeta(::MSU.System.Serialization.SerializationMetaData);
```
Enduriel, line 38: As mentioned above based on @LordMidas's suggestion I now think this should be optionally behind a flag.

msu/systems/serialization/serialization_types/table_data.nut
```diff
@@ -0,0 +1,32 @@
+::MSU.Serialization.Class.TableData <- class extends ::MSU.Serialization.Class.ArrayData
+{
+	__DataArray = null;
```
Enduriel, line 3: This is redundant

msu/systems/serialization/serialization_types/array_data.nut
LordMidas: It's used in `serde_emulator` to check the validity of the data during `__readData`

msu/systems/serialization/serialization_types/serialization_data.nut
LordMidas: We decided to not do this - and instead enforce that the metadata is always the correct metadata from the game. If people want custom things, they can store them in custom data containers.

msu/systems/serialization/metadata_emulator.nut
LordMidas: We decided to not do this - and instead enforce that the metadata is always the correct metadata from the game. If people want custom things, they can store them in custom data containers.

msu/hooks/states/world_state.nut
LordMidas: Players are serialized before because roster etc. is done from the exe before all this. But are we sure there is no BB class that is serialized after world_state?

Enduriel: I am reasonably sure, worst case we get a bug and fix it.

msu/systems/serialization/metadata_emulator.nut
```diff
@@ -40,56 +50,186 @@
...
+		if (_key in this.Data || this.RealMetaData == null)
+			return this.__getValue(_key);
+		return this.RealMetaData.getFloat(_key);
 	}
 
-	function setString( _key, _value )
+	function setFloat( _key, _value )
 	{
-		::MSU.requireString(_value);
+		::MSU.requireOneFromTypes(["float", "integer"], _value);
```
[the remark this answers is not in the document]
Enduriel: Not necessarily and we decided probably worse because of the way vanilla handles this situation

PR #344, opened 2024-04-11
feat: add functions to get actors adjacent and with specified distance

### LINKED ISSUE #343: [REQUEST] Add new getAdjacentActors function for actor.nut
Several skills and perks work off of adjacent entities, meaning anyone standing on the 6 tiles around an entity/position.
In Vanilla this is very copy&pasty done with a for-loop from 0 to 5 which then checks each tile via a `getNext(i)` function.

MSU currently only offers `::Tactical.Entities.getActorsByFunction` but that is not efficient when we know that we only need to check the adjacent tiles. That function instead iterates over every entity on the whole battlefield and it is also a bit harder to parse when reading the code.

I would like to have a `function getAdjacentActors()` for the `actor.nut` class which returns an array of up to 6 actor references.
Adjacency is where I draw the line already for this. Anything beyond adjacent needs to fall back on a `getActorsByFunction` call.

LordMidas:
Adds a more efficient way to get adjacent actors instead of having to iterate over all actors on the battlefield. Primarily targets issue #343. 

Usage for adjacent actors:
```squirrel
// To get all adjacent actors: 
::Tactical.Entities.getAdjacentActors(_tile);

// To get ALLIED adjacent actors:
::Tactical.Entities.getAdjacentActors(_tile).filter(@(_, a) a.isAlliedWith(myActor));

// To get Hostile adjacent actors
::Tactical.Entities.getAdjacentActors(_tile).filter(@(_, a) !a.isAlliedWith(myActor));

// To get Same Faction adjacent actors:
::Tactical.Entities.getAdjacentActors(_tile).filter(@(_, a) a.getFaction() == myActor.getFaction());

// To get ALLIED non-faction adjacent actors:
::Tactical.Entities.getAdjacentActors(_tile).filter(@(_, a) a.isAlliedWith(myActor) && a.getFaction() != myActor.getFaction());
```

### DIFF
--- a/msu/hooks/entity/tactical/tactical_entity_manager.nut
+++ b/msu/hooks/entity/tactical/tactical_entity_manager.nut
@@ -86,10 +86,17 @@
 	
 	q.getAlliedActors <- function( _faction, _tile = null, _distance = null, _atDistance = false )
 	{
-		if (_tile != null && _tile.ID == 0)
+		if (_tile != null)
 		{
-			::logError("The ID of _tile is 0 which means that the actor this tile was fetched from is not placed on map.");
-			throw ::MSU.Exception.InvalidValue(_tile);
+			if (_tile.ID == 0)
+			{
+				::logError("The ID of _tile is 0 which means that the actor this tile was fetched from is not placed on map.");
+				throw ::MSU.Exception.InvalidValue(_tile);
+			}
+			if (_distance == 1)
+			{
+				return this.getActorsWithinRange(_tile, 1, _atDistance ? 1 : 0).filter(@(_, _a) _a.isAlliedWith(_faction));
+			}
 		}
 
 		return this.getActorsByFunction(function(_actor) {
@@ -106,10 +113,17 @@
 	
 	q.getHostileActors <- function( _faction, _tile = null, _distance = null, _atDistance = false )
 	{
-		if (_tile != null && _tile.ID == 0)
+		if (_tile != null)
 		{
-			::logError("The ID of _tile is 0 which means that the actor this tile was fetched from is not placed on map.");
-			throw ::MSU.Exception.InvalidValue(_tile);
+			if (_tile.ID == 0)
+			{
+				::logError("The ID of _tile is 0 which means that the actor this tile was fetched from is not placed on map.");
+				throw ::MSU.Exception.InvalidValue(_tile);
+			}
+			if (_distance == 1)
+			{
+				return this.getActorsWithinRange(_tile, 1, _atDistance ? 1 : 0).filter(@(_, _a) !_a.isAlliedWith(_faction));
+			}
 		}
 
 		return this.getActorsByFunction(function(_actor) {
@@ -130,10 +144,17 @@
 		{
 			return clone this.getInstancesOfFaction(_faction);
 		}
-		else if (_tile.ID == 0)
+		else
 		{
-			::logError("The ID of _tile is 0 which means that the actor this tile was fetched from is not placed on map.");
-			throw ::MSU.Exception.InvalidValue(_tile);
+			if (_tile.ID == 0)
+			{
+				::logError("The ID of _tile is 0 which means that the actor this tile was fetched from is not placed on map.");
+				throw ::MSU.Exception.InvalidValue(_tile);
+			}
+			if (_distance == 1)
+			{
+				return this.getActorsWithinRange(_tile, 1, _atDistance ? 1 : 0).filter(@(_, _a) _a.getFaction() == _faction);
+			}
 		}
 				
 		local actors = this.getInstancesOfFaction(_faction);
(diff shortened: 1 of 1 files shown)


Enduriel (changes requested):
I don't actually think we should be adding a generic handler for distance here, we can optimize the distance=1 case, so let's, but otherwise we should allow modders to do it themselves, no reason to add extra code to handle this situation in my opinion.

LordMidas:
> I don't actually think we should be adding a generic handler for distance here, we can optimize the distance=1 case, so let's, but otherwise we should allow modders to do it themselves, no reason to add extra code to handle this situation in my opinion.

I don't think there's a clean way to handle the distance == 1 case without adding this extra code because the already present functions call `getActorsByFunction` which is distance agnostic.

TaroEld:
> I don't actually think we should be adding a generic handler for distance here, we can optimize the distance=1 case, so let's, but otherwise we should allow modders to do it themselves, no reason to add extra code to handle this situation in my opinion.

I'm of the opposite opinion, might as well make it useful.
```
q.getActorsWithinRange <- function( _tile, _max = 1, _min = 1)
{
	// add validation
	if (_max == 1)
	{
		local ret = [];

		if (_min == 0 &&  && _tile.IsOccupiedByActor)
			ret.push(_tile.getEntity());

		for (local i = 0; i < 6; i++)
		{
			if (!_tile.hasNextTile(i))
				continue;

			local nextTile = _tile.getNextTile(i);
			if (nextTile.IsOccupiedByActor)
				ret.push(nextTile.getEntity());
		}

		return ret;
	}
	else
	{
		return this.getActorsByFunction(function(_actor) {
			if (!_actor.isPlacedOnMap())
				return false;
			local distance = _tile.getDistanceTo(_actor.getTile());
			return (distance <= _max && distance >= _min)
		})
	}
}
q.getAlliedActorsWithinRange <- function(_actor, _tile, _max = 1, _min = 1)
{
	return this.getActorsWithinRange(_tile, _max, _min).filter(@(_, a) a.isAlliedWith(_actor));
}
q.getHostileActorsWithinRange <- function(_actor, _tile, _max = 1, _min = 1)
{
	return this.getActorsWithinRange(_tile, _max, _min).filter(@(_, a) !a.isAlliedWith(_actor));
}
q.getSameFactionActorsWithinRange <- function(_actor, _tile, _max = 1, _min = 1)
{
	return this.getActorsWithinRange(_tile, _max, _min).filter(@(_, a) a.getFaction() == _actor.getFaction());
}
q.getOtherFactionActorsWithinRange <- function(_actor, _tile, _max = 1, _min = 1)
{
	return this.getActorsWithinRange(_tile, _max, _min).filter(@(_, a) a.getFaction() != _actor.getFaction());
}
```

Enduriel:
Ok, fair, in that case I'd change reorganize the main function so that `getNeighboringActors` is separate. I don't feel great about having `_actor, _tile = null`, but to be fair I'm not really sure how to organize that differently
```squirrel
q.getActorsWithinRange <- function( _tile, _max = 1, _min = 1)
{
	// add validation
	if (_max == 1)
	{
		local ret = this.getNeighboringActors(_tile);
		if (_min == 0 && _tile.IsOccupiedByActor)
			ret.push(_tile);
		return ret;
	}
	else
	{
		return this.getActorsByFunction(function(_actor) {
			if (!_actor.isPlacedOnMap())
				return false;
			local distance = _tile.getDistanceTo(_actor.getTile());
			return (distance <= _max && distance >= _min)
		})
	}
}

q.getNeighboringActors <- function ( _tile ) {
	local ret = [];
	for (local i = 0; i < 6; i++)
	{
		if (!_tile.hasNextTile(i))
			continue;

		local nextTile = _tile.getNextTile(i);
		if (nextTile.IsOccupiedByActor)
			ret.push(nextTile.getEntity());
	}
	return ret;
}
q.getAlliedActorsWithinRange <- function( _tile , _max = 1, _min = 1)
{
	return this.getActorsWithinRange(_tile == null ? _actor.getTile() : _tile, _max, _min).filter(@(_, a) a.isAlliedWith(_actor));
}
q.getHostileActorsWithinRange <- function(_actor, _tile = null, _max = 1, _min = 1)
{
	return this.getActorsWithinRange(_tile == null ? _actor.getTile() : _tile, _max, _min).filter(@(_, a) !a.isAlliedWith(_actor));
}
q.getSameFactionActorsWithinRange <- function(_actor, _tile = null, _max = 1, _min = 1)
{
	return this.getActorsWithinRange(_tile == null ? _actor.getTile() : _tile, _max, _min).filter(@(_, a) a.getFaction() == _actor.getFaction());
}
q.getOtherFactionActorsWithinRange <- function(_actor, _tile = null, _max = 1, _min = 1)
{
	return this.getActorsWithinRange(_tile == null ? _actor.getTile() : _tile, _max, _min).filter(@(_, a) a.getFaction() != _actor.getFaction());
}
```

TaroEld:
getAlliedActorsWithinRange lacks the _actor parameter, otherwise LGTM. Also yes, I understand, but I think this still provides the path of least resistance for most applications.

Darxo:
I prefer a version where it is apparent from the function name, that it gathers the actors at a distance of 1 tile, like `q.getNeighboringActors <- function ( _tile )` from enduriels post. 
Otherwise it will look in code like this: `[...].getActorsWithinRange(actor.getTile())`
And it's not clear **what** range that exactly is for someone unfamiliar with that msu function.

TaroEld:
With Endys version, we'll have the best of both worlds, the clearly specific neighbours and the more generic range option.

LordMidas:
I have updated the PR with the improved functionality of `within range`. I don't think we need to add individual functions for each faction type - that can be easily done by the modder by using `.filter` on the output of these two functions `getActorsWithinRange` and `getAdjacentActors`.

Enduriel (approved):

TaroEld:
I'd not add the function parameter.

TaroEld (approved):

TaroEld:
> There is `Tactical.queryActorsInRange()` already, why not use that?
Yes, we too realized that, I think the final conclusion was 'ok we dont need this'.

LordMidas:
Yeah these were created before we realized that the vanilla function exists. However, we still need to optimize our existing functions especially for the case of adjacent actors. Also a new function to get adjacent actors like `getAdjacentActors` is a compelling addition as that is a very commonly accessed situation.

Also, the vanilla function is slower than our implementation of `getActorsByFunction` (especially when there's fewer actors on the battlefield). But that's a minor point. Nevertheless we need to optimize our existing functions for the adjacency case.

Suor:
Vanilla function lacks usability too.

PR #348, opened 2024-04-25
Continuous keybinds

TaroEld:
Adds a feature to continuously press a keybind, including trigger delays.


LordMidas (approved):
No idea about this one so you and Enduriel decide.

PR #353, opened 2024-05-05
Feat set class

LordMidas:
It represents an unordered collection of uniques. It is basically a wrapper around a table with a more convenient front-end for adding/removing items.


Suor:
Sets are usually needed to unite, intersect, subtract stuff

PR #355, opened 2024-05-10
fix: potentially pushing a skill multiple times for scheduled changes

LordMidas:
This wasn't really causing any issues currently because the scheduled changes are cleared the first time they are run, but for complete sanity this check should be baked in.


Enduriel (approved):

TaroEld (approved):
I'm often a bit uneasy with using the reference for checks, but I suppose there's no such thing as a skill uid.

PR #358, opened 2024-05-10
feat: add round function in math utilities to round to nearest multiple

LordMidas:
Should we call it `round` or something else e.g. `roundToMultiple` or `roundToNearest`?

### DIFF
--- a/msu/utils/math.nut
+++ b/msu/utils/math.nut
@@ -39,4 +39,18 @@
 		local magnitude = ::pow(10, _significantFigures - d);
 		return ::Math.round(_num * magnitude) / magnitude;
 	}
+
+	function roundToMult( _num, _multiple )
+	{
+		if (_multiple <= 0)
+		{
+			::logError("_multiple must be greater than 0");
+			throw ::MSU.Exception.InvalidValue(_multiple);
+		}
+
+		local num = ::fabs(_num);
+		local rem = num % _multiple;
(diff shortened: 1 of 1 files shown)


LordMidas:
Updated the PR with a more accurate algorithm from Taro.

msu/utils/math.nut
```diff
@@ -28,4 +28,18 @@
...
+		if (_multiple <= 0)
+		{
+			::logError("_multiple must be greater than 0");
+			throw ::MSU.Exception.InvalidValue(_multiple);
+		}
+
+		local num = fabs(_num);
+	    local rem = num % _multiple;
+	    local ret = rem < _multiple * 0.5 ? num - rem : num + _multiple - rem;
+	    return _num < 0 ? -ret : ret;
```
Enduriel, line 43: Seems to be an indentation issue here, probably mixed spaces/tabs

Enduriel (changes requested):

msu/utils/math.nut
TaroEld: Opps didnt mean to resolve

LordMidas: Thanks, fixed.

TaroEld:
> I think we can just maximally explicit and call it `roundToNearestMultiple`

I don't see an issue with that, yeah

Enduriel (approved):

TaroEld (approved):

PR #359, opened 2024-05-22
fix: costs preview not calling the original early enough

LordMidas:
The entire point of the switcheroo on JSHandle is so that the original function is called before our skill container event (onAffordablePreview) so that the values set by the original function happen before our event.

This is how it should have been in #313 and how it was in the first implementation of that PR. But in all the refactors switching to switcherooing the JSHandle, it got messed up. Here it is fixed now.


TaroEld:
Uhm, you might want to test that system extensively, then. I haven't really paid attention in detail what you've been doing, but I'd rather not wreck vanilla.

Enduriel (approved):

TaroEld (approved):

PR #360, opened 2024-06-03
feat: add deepClone function

Enduriel:
Adds a deepClone function which should really be in the standard library or at the very least in MSU for a while, this is an old PR that we overlooked and I'd like for it to be included in 1.3 if possible.

### DIFF
--- a/msu/utils/globals.nut
+++ b/msu/utils/globals.nut
@@ -127,3 +127,21 @@
 		_object = _object.get();
 	return typeof _object == "table" && "_release_hook_DO_NOT_delete_it_" in _object;
 }
+
+::MSU.deepClone <- function( _object )
+{
+	local ret;
+	if (typeof _object == "table" || typeof _object == "array")
+	{
+		foreach (key, value in _object)
+		{
+			if (typeof _object == "table") ret[key] <- ::MSU.deepClone(value);
+			else ret.push(::MSU.deepClone(value));
+		}
+	}
+	else
+	{
+		ret = clone _object;
+	}
+	return ret;
+}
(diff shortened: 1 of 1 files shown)


msu/utils/globals.nut
```diff
@@ -127,3 +127,28 @@
...
+			{
+				ret[key] <- ::MSU.deepClone(value);
+			}
+			return ret;
+		case "array":
+			ret = array(_object.len());
+			foreach (idx, value in _object)
+			{
+				ret[idx] = ::MSU.deepClone(value);
+			}
```
LordMidas, line 148: Use native map function: `return _object.map(@(_a) ::MSU.deepClone(_a));`

msu/utils/globals.nut
```diff
@@ -127,3 +127,28 @@
...
+::MSU.deepClone <- function( _object )
+{
+	local ret;
+	switch (typeof _object)
+	{
+		case "table":
+			ret = {};
+			foreach (key, value in _object)
+			{
+				ret[key] <- ::MSU.deepClone(value);
```
LordMidas, line 140: How does cloning of tables handle delegation? Ideally we'd want a deep clone of a table with tables which has delegation set up should return a setup with the exact same delegation setup.

LordMidas (changes requested):

msu/utils/globals.nut
Enduriel: This is a good point and I agree

Enduriel:
huh apparently the reforged branch already has a `deepClone` but it is less nice and this one is generally superior, especially with the delegation cloning.

LordMidas (approved):
Please edit the PR description to say `deepClone` instead of `deepEquals`.

TaroEld (approved):
I wonder if it works with BB objects.

PR #365, opened 2024-06-09
refactor: change getFiles to return an array of filenames

Enduriel:


TaroEld (approved):

msu/systems/persistent_data/persistent_data_mod_addon.nut
```diff
@@ -17,9 +17,14 @@
 	}
 
 	// MSU 1.3.0
+	function __prefixRaw()
```
LordMidas, line 20: This only returns something so it should be `__getFilenamePrefix` and the other function which actually applies it should be `__prefixFilename`.

LordMidas (changes requested):

msu/systems/persistent_data/persistent_data_mod_addon.nut
```diff
@@ -17,9 +17,14 @@
 	}
 
 	// MSU 1.3.0
+	function __getFilenamePrefix()
+	{
+		return format("%s%s#", ::MSU.System.PersistentData.FilePrefix, this.Mod.getID());
+	}
+
 	function __prefixFileName( _fileName )
```
LordMidas, line 25: Decapitalize `N` then looks good

LordMidas (approved):

PR #366, opened 2024-06-10
refactor: move classes from Serialization.Class to Class

Enduriel:
this is for consistency with the rest of the codebase. If we're okay with this I thought we should have a PR that we could approve so that if it's fine we can just merge it in.


TaroEld (approved):
I mean I suppose you just used regex or sth so I'm not gonna check every entry

LordMidas (approved):
- Remove the `Class` key from `::MSU.Serialization` on line 2. 
- If you have replaced all occurrences then looks good to merge.

PR #367, opened 2024-06-10
feat: add deepEquals function

Enduriel:
I'd like this in 1.3 because
- we need something like this just like deepClone and it's a little embarrassing we don't have it
- it would make doc a lot easier to write when talking about comparing reserialized data

### DIFF
--- a/msu/utils/globals.nut
+++ b/msu/utils/globals.nut
@@ -148,3 +148,32 @@
 			return _object;
 	}
 }
+
+::MSU.deepEquals <- function(_a, _b)
+{
+	if (_a == _b)
+		return true;
+	if (typeof _a != typeof _b)
+		return false;
+	if (typeof _a != "array" && typeof _a != "table")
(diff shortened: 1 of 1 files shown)


msu/utils/globals.nut
```diff
@@ -148,3 +148,32 @@
 			return _object;
 	}
 }
+
+::MSU.deepEquals <- function(_a, _b)
```
LordMidas, line 152: The original PR for this #339 was closed. I left [a comment](https://github.com/MSUTeam/MSU/pull/339#discussion_r1506845018) there that I think still applies. Please check reassess.

LordMidas (changes requested):

msu/utils/globals.nut
Enduriel: Alright, thanks, will adjust based on that feedback.

LordMidas (approved):

PR #368, opened 2024-06-10
refactor: add back VanillaID due to legends using it

Enduriel:
Annoying, but better than having to deal with an incompatible legends version.


LordMidas (approved):
Legends should have `L` capitalized (nitpick but sorry :( ). Looks good to merge after that change.

PR #369, opened 2024-06-10
fix: handling for adding/removing weapontype during hooks

Enduriel:


TaroEld (approved):
works for me

msu/hooks/items/weapons/weapon.nut
```diff
@@ -92,6 +88,7 @@
 
 	q.addWeaponType <- function( _weaponType, _setupCategories = true )
 	{
+		this.setupWeaponType(false);
```
LordMidas: I don't think this works. Example:
- A vanilla weapon that does `this.m.Categories = "Axe"` in its `create()`.
- Mod A hooks this function, calls the `__original()` and does `addWeaponType(Spear)`.
- This goes to `setupWeaponType(false)` which sets the WeaponType to `None`. Then goes through all the categories, sets it to `Axe`. Then the `addWeaponType` function set it to `Axe | Spear`. Note that the Categories are not updated/set here.
- Mod B hooks afterward and says `addWeaponType(Hammer)`.
- This goes to `setupWeaponType(false)` which sets the WeaponType to `None`. Then goes through all the categories, sets it to `Axe`. Then the `addWeaponType` function set it to `Axe | Hammer`. Note that this means that the `Spear` added by Mod A is lost.

I guess the solution is to also set up the categories immediately after `setupWeaponType` in `addWeaponType` and `removeWeaponType`.

msu/hooks/items/weapons/weapon.nut
LordMidas, line 1: I don't think the Boolean is necessary. We can simply always update both the weapontype and categories before adding/removing weapontypes. I've pushed a commit that you can check.

LordMidas (changes requested):

msu/hooks/items/weapons/weapon.nut
LordMidas: Ah, I guess one problem there is that currently we allow adding weapon types without updating categories `_setupCategories = true` by default which can be passed as `false`. So "technically" it allows someone to add a weapon type to a weapon but not list that in the categories. Although this probably shouldn't be allowed - I don't remember why I added that Boolean param, maybe it was for performance reasons if someone wants to add/remove several weapon types from a weapon they don't update categories every single time. But in hindsight this seems like an unnecessary performance optimization and allows a certain design choice which is bad and shouldn't be allowed i.e. weapon types added without showing in categories.

In any case, the thing is that the new approach will no longer allow that design choice to work because will always ensure that every weapon type shows up in the categories. Are we ok with this side-effect?

LordMidas (approved):

PR #371, opened 2024-06-13
fix: table and class cases in deepEquals

LordMidas:
- For the table case add check for key existing in _b.
- For the instance case require that the class of both instances is the same.

### DIFF
--- a/msu/utils/globals.nut
+++ b/msu/utils/globals.nut
@@ -162,7 +162,7 @@
 				return false;
 			foreach (k, v in _a)
 			{
-				if (!this.deepEquals(v, _b[k]))
+				if (!(k in _b) || !this.deepEquals(v, _b[k]))
 					return false;
 			}
 			return true;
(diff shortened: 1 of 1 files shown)


msu/utils/globals.nut
```diff
@@ -176,11 +176,11 @@
 			}
 			return true;
 		case "instance":
-			if (!(_b instanceof _a.getclass()))
+			if (!(typeof _b != "instance") || _a.getclass() != _b.getclass())
 				return false;
-			foreach (k, v in _a.getclass())
+			foreach (k, v in _a)
```
Enduriel, line 181: why is this no longer getclass? afaik this has to be getclass otherwise the instances `_next` metamethod is used.

Enduriel (changes requested):

msu/utils/globals.nut
LordMidas: I see. I didn't account for the `_nexti` thing. Does the `_nexti` overwrite not apply to iteration on classes?

Enduriel: No

LordMidas: Ok I'll change it to `getClass()`

Enduriel (approved):

TaroEld (approved):

PR #379, opened 2024-07-01
fix: being forced to have a non-vanilla ui due to font fixes

Enduriel:


TaroEld (approved):

LordMidas (approved):
Does this add `mb` and `mi` tags for bbcode processing? If yes, then imo that should be separated into a feat commit. Otherwise looks good.

PR #381, opened 2024-07-02
feat: add small popup notification and expose functions for other mods to add messages

### LINKED ISSUE #373: [REQUEST] Allow selective unsubscribing from update notifications
MSU allows authors to link their mod with a website in order to produce update notifications for a player when a newer version was released.
However not everyone wants to update to the newer version.
For example:
- Newer version breaks other mods
- Newer version is not save-game compatible
- Newer version changes balance, or adds/removes features, that the user does not agree with

- Add a mod settings page which is only visible while in the main menu and uses the persistent data storage.
- This settings page contains a boolean setting for every currently registered mods which use an update source.
- By default those settings are enabled
- A mod, that is disabled here, will no longer prompts you to visit the source for the latest version

### Need for discussion
How to treat disabled update notifications for mods which are removed?

**Team: auto-cleanup:**
- A player who at one point purposefully played an old version of a mod might have removed it later. Then one day he downloads the latest version of it as he now wants to try the latest changes. But at this point he never remembers to enable the update notification again
- A streamer who plays different modlists in parallel does not care much about the update notification of the older version of one of his mods. It is more important to him, that the other playthrough with the most recent version of the same mod actually notifies him about updates.
**Team: keep forever**
- Only thing against that ruling are players (mostly modders) who try out various different mod setups for bug testing. Once they test something with a minimal mod setup, they will wipe every setting and have to re-do those

Maybe the answer here is to introduce an additional boolean setting for **Wipe notification setting when mod not present**

TaroEld:
This comes with a refactor of the basic popup functions.
The popup now has 3 'states': None, Small and Full
None means hidden.
Small means a small popup icon will be present, with the MSU icon overlaid with an exclamation mark, and optionally a footer message.
The user can then click on this small popup to bring up the big popup. This one has been improved to be more modular, and allows other mods to post messages via the mod addon API.


msu/systems/debug/debug_mod_addon.nut
```diff
@@ -39,4 +39,13 @@
 	{
 		::MSU.System.Debug.print(_text, this.Mod.getID(), ::MSU.System.Debug.LogType.Error, _flagID);
 	}
+
+	function addMessage( _text )
```
LordMidas, line 43: Would prefer the name `addPopupMessage` here.

msu/systems/debug/debug_mod_addon.nut
```diff
@@ -39,4 +39,13 @@
...
+	{
+		::MSU.System.Debug.addPopupMessage({
+			modID = this.Mod.getID(),
+			modName = this.Mod.getName(),
+			text = _text
+		})
```
LordMidas, line 49: `;`

msu/systems/debug/debug_system.nut
```diff
@@ -119,4 +119,10 @@
 			}
 		}
 	}
+
+	function addPopupMessage(_info)
```
LordMidas, line 123: `( _info )`

scripts/mods/msu/popup.nut
```diff
@@ -3,8 +3,13 @@ this.popup <- {
 		Visible = false,
 		Animating = false,
 		JSHandle = null,
-		TextCache = "",
-		ForceQuit = false
+		States = {
```
LordMidas, line 6: As this is an "enum" it should be singular imo i.e. `State = `

scripts/mods/msu/popup.nut
```diff
@@ -17,32 +22,65 @@ this.popup <- {
 		return this.m.Animating;
 	}
 
-	function showRawText( _text, _forceQuit = false )
+	function addMessage(_info)
```
LordMidas, line 25: `( _info )`

ui/mods/msu/utilities.js
```diff
@@ -1,6 +1,7 @@
 var MSU = {
 	ID : "mod_msu",
 	Hooks : {},
+	Popup : Screens.MSUPopup
```
LordMidas, line 4: What is this for?

msu/systems/debug/debug_system.nut
```diff
@@ -119,4 +119,10 @@
...
+	{
+		::MSU.Popup.addMessage(_info);
+		::MSU.Popup.setInfoText(_info.modID);
```
LordMidas, line 126: What's the difference between the message and the info text and where is each shown in the UI? I assume the message is the entry in the popup itself.

LordMidas (changes requested):
You can ignore the style-related review comments for now - I just put them there for completion sake for later. Just focus on the functionality/code related comments. Overall seems like a reasonable set of changes to me but with some questions/suggestions.

ui/mods/msu/utilities.js
TaroEld: It's just easier to remember to use MSU.Popup over Screens.MSUPopup.

msu/systems/debug/debug_system.nut
TaroEld: Info text is below the item
image

message is in the popup
image

scripts/mods/msu/popup.nut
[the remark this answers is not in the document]
TaroEld: I'm not sure if we should give modders that power.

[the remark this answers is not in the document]
Enduriel: FWIW I do, but I'm okay with that being an explicit call to a separate function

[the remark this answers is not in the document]
TaroEld: Frankly I can't think of a game where a mod can just force you to close it (without crashing).
I'd prefer to just leave it in the popup.nut file for now.

gfx/mods/msu/buttons/mod_update.png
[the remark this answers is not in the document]
TaroEld: Unfortunately that looks like pixelated excrement when scaled down.

scripts/mods/msu/popup.nut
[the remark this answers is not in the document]
LordMidas: I'm ok with it being a separate explicit function call.

msu/squirrel_hooks/mod_hooks.nut
Enduriel, line 1: What is this doing here?

gfx/mods/msu/buttons/mod_update.png
Enduriel, line 1: Ideally, we should use an svg imo, I made a clone of this for the MSU launcher [here](https://github.com/MSUTeam/MSU-Launcher/blob/main/assets/assets/gfx/icons/warning.svg)

scripts/mods/msu/popup.nut
```diff
@@ -1,10 +1,15 @@
...
+		Full = 2
+	}
 	m = {
 		Visible = false,
 		Animating = false,
 		JSHandle = null,
-		TextCache = "",
-		ForceQuit = false
+		ForceQuit = false,
+		FunctionBuffer = [],
```
Enduriel, line 12: Yeah I'm not sure why I called it a cache, buffer is the correct term, I like the change here.

ui/mods/msu/css/popup.css
```diff
@@ -1,4 +1,5 @@
 .msu-popup {
+	display: none;
 	height: 72.0rem !important;
```
Enduriel, line 3: Not a huge fan of `display: none;` being part of the CSS class, I think this should be a separate thing we flick on/off rather than a part of the class we override.

Enduriel (changes requested):
Overall great, some nitpicks

msu/squirrel_hooks/mod_hooks.nut
TaroEld: As you can hide the small popup by rightclicking it, you need a way to bring it up again.

ui/mods/msu/css/popup.css
TaroEld: All jquery hide() and show() does is toggle between None and whatever the last display was. I can remove it, but then I'll need to add a setState(None) somewhere during setup, it'll be the same thing.

msu/squirrel_hooks/mod_hooks.nut
Enduriel: It's just an empty file?

ui/mods/msu/css/popup.css
Enduriel: I would prefer that, this is a conceptual thing, the popup style isn't to be invisible, you can hide it, but that's not part of its class.

msu/squirrel_hooks/mod_hooks.nut
TaroEld: Huh the comment showed on the keybind for me I think. Idk ill look

gfx/mods/msu/buttons/mod_update.png
TaroEld: Didn't work, and not really necessary for the small version here.

ui/mods/msu/css/popup.css
TaroEld: Was deemed too annoying to do.

scripts/mods/msu/popup.nut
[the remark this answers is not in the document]
TaroEld: I'll leave the decision to the next version.

Enduriel (approved):

LordMidas (approved):

PR #383, opened 2024-07-05
feat: auto calculate DamageType ratio in skills

LordMidas:
Originally we calculated approximate ratios for these "mixed" damage type skills based on the number of injuries there are in vanilla and hard-coded them as weights in the containers so that the injury inflicting behavior with MSU damage type system corresponds to the probability of injuries in vanilla. However, mods may change the number of different types of injuries, causing this ratio to no longer represent the new ratio of injuries. This PR ensures that the ratio is calculated dynamically.

I have used the MSU.Math.roundToMult function #358 to round it to the nearest multiple of 5. For the case of vanilla this only affects `BluntAndPiercing` which in vanilla was giving a ratio of 58/42 but we had it set at 45/55. The others were already exact at multiples of 5. I believe rounded to the nearest multiple of 5 is still good as it keeps the damage type ratios more "clean" for the players in tooltips.

### DIFF
--- a/msu/classes/weighted_container_damage_type.nut
+++ b/msu/classes/weighted_container_damage_type.nut
@@ -45,7 +45,7 @@
 					local piercing = ::Const.Injury.PiercingBody.len() + ::Const.Injury.PiercingHead.len();
 					local total = (blunt + piercing).tofloat();
 					this.add(::Const.Damage.DamageType.Blunt, ::MSU.Math.roundToMult(100 * blunt / total, 5)); // We round to the nearest multiple of 5 to make it "cleaner" for players
-					this.add(::Const.Damage.DamageType.Piercing, ::MSU.Math.roundToMult(100 * blunt / total, 5));
+					this.add(::Const.Damage.DamageType.Piercing, ::MSU.Math.roundToMult(100 * piercing / total, 5));
 					break;
 
 				case ::Const.Injury.BurningAndPiercingBody:
--- a/msu/classes/weighted_container_damage_type.nut
+++ b/msu/classes/weighted_container_damage_type.nut
@@ -41,18 +41,27 @@
 					break;
 
 				case ::Const.Injury.BluntAndPiercingBody:
-					this.add(::Const.Damage.DamageType.Blunt, 55);
-					this.add(::Const.Damage.DamageType.Piercing, 45);
+					local blunt = ::Const.Injury.BluntBody.len() + ::Const.Injury.BluntHead.len();
+					local piercing = ::Const.Injury.PiercingBody.len() + ::Const.Injury.PiercingHead.len();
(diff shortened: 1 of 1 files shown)


msu/hooks/skills/skill.nut
```diff
@@ -4,10 +4,11 @@
 
 	q.create = @(__original) function()
 	{
-		__original();
 		if (this.m.DamageType == null)
 			this.m.DamageType = ::MSU.Class.WeightedContainer();
 
+		__original();
+
```
TaroEld, line 11: Is this order change intended?

msu/hooks/skills/skill.nut
```diff
@@ -33,18 +34,27 @@
 					break;
 
 				case ::Const.Injury.BluntAndPiercingBody:
-					this.m.DamageType.add(::Const.Damage.DamageType.Blunt, 55);
-					this.m.DamageType.add(::Const.Damage.DamageType.Piercing, 45);
+					local blunt = ::Const.Injury.BluntBody.len() + ::Const.Injury.BluntHead.len();
+					local piercing = ::Const.Injury.PiercingBody.len() + ::Const.Injury.PiercingHead.len();
+					local total = (blunt + piercing).tofloat();
+					this.m.DamageType.add(::Const.Damage.DamageType.Blunt, ::MSU.Math.round(100 * blunt / total, 5)); // We round to the nearest multiple of 5 to make it "cleaner" for players
+					this.m.DamageType.add(::Const.Damage.DamageType.Piercing, ::MSU.Math.round(100 * blunt / total, 5));
```
TaroEld, line 41: Should be `::MSU.Math.round(100 * piercing/ total, 5))`, no?

TaroEld (commented):
What about all the other combinations of damage types? Are other mods supposed to add to this then?

msu/hooks/skills/skill.nut
LordMidas: Yes that's part of a separate fix commit. It makes sense to me that we initialize it before calling the original function otherwise someone may hook the original function and try to interact with the DamageType field but it's null at that time.

LordMidas:
> What about all the other combinations of damage types? Are other mods supposed to add to this then?

This is meant to convert vanilla skills (or modded skills which use vanilla injuries but don't use the MSU damage type system) to the MSU damage type system. So we cover all the vanilla injury combinations.

msu/hooks/skills/skill.nut
LordMidas: Yes, will fix

TaroEld:
So this should be roundSig then

TaroEld (approved):

Enduriel (approved):

PR #384, opened 2024-07-05
fix: prevent adding reserved keywords to ordered map

LordMidas:


Enduriel (approved):
I don't think this will correctly handle keys like weakref but is probably the best we can do

TaroEld (approved):

PR #386, opened 2024-07-05
feat: add function in math utils to round to significant figures

LordMidas:


msu/utils/math.nut
```diff
@@ -28,4 +28,15 @@
...
 		return _min + (::Math.rand(0, 2147483647) / 2147483647.0) * (_max - _min);
 	}
+
+	function roundSig( _num, _significantFigures )
+	{
+		::MSU.requireInt(_significantFigures);
+		if (_num == 0)
+			return 0;
+
+		local d = ceil(log10(_num < 0 ? -_num : _num));
```
Enduriel, line 38: use `::Math.abs` instead

msu/utils/math.nut
```diff
@@ -28,4 +28,15 @@
...
+		local magnitude = pow(10, _significantFigures - d.tointeger()); // tointeger to prevent floating point accuracy issues
```
Enduriel, line 39: `tointeger()` should be on the previous line I think

Enduriel (changes requested):

TaroEld:
What endy said

msu/utils/math.nut
LordMidas: I've use `fabs` which is a native function and hopefully faster/more efficient than `::Math.abs`.

TaroEld (approved):

Enduriel (approved):

PR #387, opened 2024-07-07
refactor: temporarily print full stack trace when deserialization errors

Enduriel:
This is to help diagnose the log spam we are getting from some people


TaroEld (approved):

LordMidas (approved):
I pushed a commit adding a comment to the code line to mark it as temporary as well. You can squash that.

Minor point: imo the commit message shouldn't say `temporarily`. Technically speaking commit messages are supposed to reflect what the code change does, not necessarily the long term intention behind the change. That information can go in the commit description. So a proper message here could be:

"
refactor: print full stack trace when deserialization errors

This change should be removed once the log spam issues are resolved.
"

PR #389, opened 2024-07-11
fix: handling for accessing skill DamageType before hookTree

LordMidas:
Add a new class for DamageType that extends WeightedContainer and overwrites all functions to first initialize the DamageType of the skill before returning the base function. This ensures that hooks on the skill's create function before hookTree that try to add/remove a DamageType or try to access any other value related to the skill's DamageType do not break the auto-assigning of DamageTypes for skills that don't come with DamageType defined (e.g. vanilla skills).


Enduriel (approved):
Assuming this is tested, lgtm, though it's annoying that it has to be done this way.

TaroEld (approved):
I havent tested this so I hope you did.

PR #391, opened 2024-07-11
Fix handling for ref type defparams in addEvent skills function

LordMidas:
- Handling for null defparam is a bugfix and should be added.
- Other ref-type params are something we can't deal with (without adding a custom wrapper with vargv, which I don't want to do here) so we simply disallow the usage of this function to add those kinds of functions. They can be added by modders in the old school manual way to add skill events.

### DIFF
--- a/msu/utils/skills.nut
+++ b/msu/utils/skills.nut
@@ -39,6 +39,9 @@
 				{
 					foreach (i, defparam in info.defparams)
 					{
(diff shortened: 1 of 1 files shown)


Enduriel (approved):

Enduriel:
I thing we may wish to consider vargv handling as a fallback in these situations

PR #395, opened 2024-07-14
fix: vanilla bb code toggle not applying to text bolded using JS

Enduriel:
Went for a different approach with this one which is imo cleaner, previously only text bolded using bbcode would be unbolded, but it turns out vanilla actually bolds some text directly with JS, this is able to handle all those cases. I had to exclude title fonts as those are always bold.


LordMidas (approved):
I don't understand most of these shenanigans so I approve.

TaroEld (approved):

PR #396, opened 2024-07-15
Fix serialization metadata becoming invalid

LordMidas:


Enduriel (approved):
Actually ultimately quite a clean solution even if the doubled `onBeforeSerialize` scares me a bit, I feel like this might be able to cause issues but all the scenarios I could come up with are covered, a more 'correct' approach would be to use a wrapper around the real metadata during `onBeforeSerialize` which duplicates everything into an emulator, but that might not be worth the work as this is quite clean

TaroEld (approved):

PR #397, opened 2024-07-15
refactor: streamline certain load files to include entire folders

LordMidas:
So that MSU modules outside the MSU zip which are trying to add or modify classes in these folders are able to do that before the classes are instantiated.

### DIFF
--- a/msu/vanilla_mod/load.nut
+++ b/msu/vanilla_mod/load.nut
@@ -3,6 +3,6 @@ local function includeFile( _file )
 	::MSU.includeFile("msu/vanilla_mod/", _file + ".nut");
 }
 includeFile("setup_vanilla_mod");
-includeFile("vanilla_keybinds");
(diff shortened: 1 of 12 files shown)


msu/vanilla_mod/load.nut
```diff
@@ -3,6 +3,6 @@ local function includeFile( _file )
 	::MSU.includeFile("msu/vanilla_mod/", _file + ".nut");
 }
 includeFile("setup_vanilla_mod");
```
Enduriel, line 5: remove this includeFile if it's only being used once imo

TaroEld (approved):

Enduriel (approved):

msu/vanilla_mod/load.nut
LordMidas: Eh, doesn't matter imo. I just copied the setup as in other load files and it's ok e.g. if we wanna extend this file with other files in the future.

PR #398, opened 2024-07-16
feat: add utility function for benchmarking

LordMidas:
This is a useful function for benchmarking a single function or comparing the benchmarks of several functions. Usage:

```squirrel
// Let's say we have 3 functions called foo, bar, baz
// we pass them as references in an array and assign them an id that will be printed in log
::MSU.Utils.benchmark([
	["Endy",  foo],
	["Midas", bar],
	["Taro", baz]
]);
```
This will print to log the performance of the 3 functions as well as their relative performance to the first function i.e. in this case `Endy`.

In case your functions need parameters you can manually call them via passing lambdas to the utility function and using local variables as the args:
```squirrel
::MSU.Utils.benchmark([
	["Endy",  @() foo(pass args here)],
	["Midas", @() bar(pass args here)],
	["Taro", @() baz(pass args here)]
]);
```

Sample:
image

### DIFF
--- a/msu/utils/utils.nut
+++ b/msu/utils/utils.nut
@@ -231,6 +231,7 @@
 			};
 		}
 
+		::logWarning("MSU -- Benchmark with " + _iterations + " iterations");
 		foreach (i, entry in times)
 		{
 			local diff = entry.TimePerIteration - times[0].TimePerIteration;
--- a/msu/utils/utils.nut
+++ b/msu/utils/utils.nut
@@ -236,11 +236,11 @@
 			local diff = entry.TimePerIteration - times[0].TimePerIteration;
 			local color = i == 0 ? "" : (diff <= 0 ? "green" : "red"); // the base function is printed without color, faster in green, slower in red
 
-			local timePerIterationText = format("<div style='display: inline-block; color:%s;'>%f (%s%f)</div>", color, entry.TimePerIteration, diff >= 0 ? "+" : "", diff);
+			local timePerIterationText = format("<span style='color:%s;'>%f (%s%f)</span>", color, entry.TimePerIteration, diff >= 0 ? "+" : "", diff);
 			diff = entry.TotalTime - times[0].TotalTime;
 			local totalTimeText = entry.TotalTime + (diff < 0 ? " (" : " (+") + diff + ")";
-			local timePctText = format("<div style='display: inline-block; color:%s'>%i%% %s</div>", color, ::Math.abs(100 * diff / times[0].TotalTime), diff < 0 ? "faster" : "slower");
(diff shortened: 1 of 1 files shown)


TaroEld (approved):

msu/utils/utils.nut
```diff
@@ -203,6 +203,48 @@
...
+
+			local totalTime = timer.silentGet() - timeStart;
+
+			times[i] = {
+				TotalTime = totalTime,
+				TimePerIteration = totalTime / _iterations
+			};
+		}
+
+		::logWarning("MSU -- Benchmark with " + _iterations + " iterations");
```
Enduriel, line 234: logInfo imo, this should't be a warning

msu/utils/utils.nut
```diff
@@ -203,6 +203,48 @@
...
+				f();
+			}
+
+			local totalTime = timer.silentGet() - timeStart;
+
+			times[i] = {
+				TotalTime = totalTime,
+				TimePerIteration = totalTime / _iterations
+			};
+		}
```
Enduriel, line 232: I don't think this is how we should be using the timer, just create it and `silentStop` once we're done, that way it handles the subtraction rather than this function.

Enduriel (changes requested):

msu/utils/utils.nut
```diff
@@ -203,6 +203,48 @@
...
+		foreach (i, entry in times)
+		{
+			local diff = entry.TimePerIteration - times[0].TimePerIteration;
+			local color = i == 0 ? "" : (diff <= 0 ? "green" : "red"); // the base function is printed without color, faster in green, slower in red
+
+			local timePerIterationText = format("<div style='display: inline-block; color:%s;'>%f (%s%f)</div>", color, entry.TimePerIteration, diff >= 0 ? "+" : "", diff);
+			diff = entry.TotalTime - times[0].TotalTime;
+			local totalTimeText = entry.TotalTime + (diff < 0 ? " (" : " (+") + diff + ")";
+			local timePctText = format("<div style='display: inline-block; color:%s'>%i%% %s</div>", color, ::Math.abs(100 * diff / times[0].TotalTime), diff < 0 ? "faster" : "slower");
+			::logInfo(format("<div style='display: inline-block;'>%s: %s ms (Total: %s ms) -- %s</div>", _functions[i][0] + "", timePerIterationText, totalTimeText, i == 0 ? "base time" : timePctText));
```
Enduriel, line 244: instead of div style inline-block, this should be a `span` element, then you can remove that part.

msu/utils/utils.nut
LordMidas: I just wanted a different color for the header. I think a different color for the header so it separates nicely from the first line (base function timing) visually. What do you think?

Enduriel (approved):

PR #399, opened 2024-07-20
feat: add removeValues function to array utilities

LordMidas:
This function removes given values from an array in place as opposed to the native filter function that returns a new array. Doing `_values.find` is expected to be faster than first converting `_values` to a local table and then using `in` for each element in `_array`. This is because looking up values that don't exist in a table is a slower operation than using `find` on an array especially at smaller len - which `_values` is expected to have.

### DIFF
--- a/msu/utils/array.nut
+++ b/msu/utils/array.nut
@@ -64,6 +64,7 @@
 				_array.remove(i);
 			}
 		}
+		return _array;
 	}
 
 	function shuffle( _array )
--- a/msu/utils/array.nut
+++ b/msu/utils/array.nut
@@ -55,6 +55,17 @@
 		if (idx != null) return _array.remove(idx);
 	}
 
+	function removeValues( _array, _values )
+	{
+		for (local i = _array.len() - 1; i >= 0; i--)
+		{
+			if (_values.find(_array[i]) != null)
+			{
+				_array.remove(i);
+			}
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


msu/utils/array.nut
```diff
@@ -55,6 +55,17 @@
...
+	function removeValues( _array, _values )
+	{
+		for (local i = _array.len() - 1; i >= 0; i--)
+		{
+			if (_values.find(_array[i]) != null)
+			{
+				_array.remove(i);
+			}
+		}
+	}
```
TaroEld, line 67: You should still return the array, its always annoying to remember which function does and which doesn't.

TaroEld (changes requested):

msu/utils/array.nut
LordMidas: Is it better to return the array itself or a new array containing all the values that were removed?

TaroEld: Returning the removed values is not consistent with other languages imo, if you need that functionality its very easy to write it yourself

Enduriel: We discussed and think this would be non-standard so no

Enduriel (approved):

TaroEld:
Pretty sure endy meant returning the removed values as an array would be non standard.

TaroEld (approved):

PR #400, opened 2024-07-20
fix: some msu exceptions in screens bb classes not throwing properly

LordMidas:


Enduriel (approved):

TaroEld (commented):
I'm not sure why it should even throw if you try to show or hide and it's not visible or already hidden

LordMidas:
> I'm not sure why it should even throw if you try to show or hide and it's not visible or already hidden

I don't know about that - that is something you and Endy should decide about. I just fixed the exception throwing.

TaroEld (approved):

PR #401, opened 2024-07-20
feat: add functions to colorize values, mults and percentages

LordMidas:


msu/utils/text.nut
```diff
@@ -28,4 +28,97 @@
...
+		}
+	}
+
+	// Returns 0.75 as red 25% and 1.75 as green 75%
+	// Used to colorize mults which are applied to other values
+	// Use case example: when someone has a 0.75 multiplier to their defense and you want to write "25% less defense" in tooltip
+	function colorizeMult( _value, _kwargs = null )
+	{
+		local kwargs = {
+			AddMoreLess = false,
```
LordMidas, line 77: Should this default to `true` or `false`? In most of the practical use cases you almost always want to append `more` or `less`. But the "purity" of the function would suggest that it should only return a colorized mult by default without additional appendages.

msu/utils/text.nut
```diff
@@ -28,4 +28,97 @@
...
+				if (!(key in kwargs)) throw "invalid parameter " + key;
+				kwargs[key] = value;
+			}
+		}
+
+		if (_value < kwargs.CompareTo)
+		{
+			if (!kwargs.AddSign && _value < 0) _value *= -1;
+			if (kwargs.AddPercent) _value = _value + "%";
+			return kwargs.Invert ? this.colorGreen(_value) : this.colorRed(_value);
```
Darxo, line 54: this should be colorPositive and colorNegative

msu/utils/text.nut
```diff
@@ -28,4 +28,97 @@
...
+		}
+
+		if (_value > kwargs.CompareTo)
+		{
+			if (kwargs.AddSign && _value > 0) _value = "+" + _value;
+			if (kwargs.AddPercent) _value = _value + "%";
+			return kwargs.Invert ? this.colorRed(_value) : this.colorGreen(_value);
```
Darxo, line 61: this should be colorNegative and colorPositive

msu/utils/text.nut
```diff
@@ -28,4 +28,97 @@
 	{
 		return this.color(::Const.UI.Color.NegativeValue, _string);
 	}
+
+	function colorizeValue( _value, _kwargs = null )
+	{
+		local kwargs = {
+			AddSign = true,
```
LordMidas, line 35: Should this default to `true` or `false`? Do you by default always want a sign attached? How has been the experience using this function in Reforged?

msu/utils/text.nut
Darxo: I like it being false. That's more intuitive for a first-time user and for modder who prefer other terms.

msu/utils/text.nut
```diff
@@ -28,4 +28,97 @@
...
+			else
+				ret += _value >= 1.0 ? " more" : " less";
+		}
+
+		return ret;
+	}
+
+	// Returns 0.75 as green 75% and 1.75 as green 175%, and -0.75 as red 75% and -1.75 as red 175%
+	// Used when showing a certain percentage of another value
+	// Use case example: when you want to say that you gain 25% of your hitpoints as stamina
```
Darxo, line 112: An alternative use-case is coloring flat armor penetration or armor damage

Darxo (changes requested):

msu/utils/text.nut
Darxo: I like it being false. That's more intuitive for a first-time user and for modder who prefer other terms.

LordMidas: An alternative would be to pass an array of preferred words e.g. `MoreLess = ["more", "less"]`. By default we could have an empty array but this could allow modders to very easily use the same function but with custom terminology.

LordMidas: I went with an alternative approach and added a separate function for colorizing mults with text. This keeps the overall thing simpler.

msu/utils/text.nut
LordMidas: I've found that in most of our use cases we didn't want the sign attached, so by default I think I'll make it `false`.

msu/utils/text.nut
```diff
@@ -28,4 +28,92 @@
 	{
 		return this.color(::Const.UI.Color.NegativeValue, _string);
 	}
+
+	function colorizeValue( _value, _kwargs = null )
+	{
+		local kwargs = {
+			AddSign = false,
+			CompareTo = 0,
+			Invert = false,
```
LordMidas, line 37: Should this be `InvertColor` or just `Invert`?

TaroEld (approved):

msu/utils/text.nut
Enduriel: `InvertColor` imo

Enduriel (changes requested):

Enduriel (approved):

PR #405, opened 2024-07-26
fix: ai paying no AP cost for pick up weapon behavior

LordMidas:
Similar to the fixes we made in #181 but we missed this behavior. I'm thinking now, this doesn't seem perfect because any modder who tries to pay action with an empty array will end up getting wrong behavior (although our [documentation](https://github.com/MSUTeam/MSU/wiki/Item-Actions) on our changes to item action stuff should make it clear that we completely change the system, so there's that).

Is there any other approach we can do to reconsider this thing from scratch? If not, then at least this fix needs to go in. Check my alternative approach in #406


Enduriel:
closed in favor of #406

PR #410, opened 2024-08-13
refactor: print specific file instead of full stack trace for invalid data

LordMidas:
Also reverts the temporary 44b46fdbdb7da6be555a0f5a52ea6b518be1aa02 as we no longer need to print the full stack trace as the "temporary" situation has been resolved. I am not aware of any current issues in that department.


Enduriel (approved):

TaroEld (approved):
Its a bit arcane but I assume its necessary

PR #411, opened 2024-08-29
fix: add create to skill and remove contains check from hookTree

LordMidas:
The contains check will cause leaf classes which do not have `create` defined to not be properly hooked by our hookTree. The solution to this is to add `create` directly to the base `skill` class and then hookTree it.


Suor:
Will break mods that old style loop parent up until there is no create, theoretically.

Practically the Spin skill from a bug does have `create()`.

PR #412, opened 2024-08-30
fix: allow integers and floats in integer serialization

LordMidas:
In vanilla ser/de functions there are cases where a float is ser/deserialized as an integer. We need to allow this behavior in our emulators otherwise things such as Combat Simulator throw exceptions when trying to ser/de players using MSU ser/de emulators. 

The changes in this PR cause there to be no exception, but we still print an error to the log for the unexpected data type.

### DIFF
--- a/msu/systems/serialization/serde_emulator.nut
+++ b/msu/systems/serialization/serde_emulator.nut
@@ -70,13 +70,13 @@
 
 		function writeI32( _int )
 		{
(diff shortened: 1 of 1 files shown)


Enduriel (approved):
I hate that this is the case but sure

LordMidas:
Closed in favor of #444. If new problems from vanilla show up, then this might need to be re-opened.

PR #416, opened 2024-09-04
Add GithubTags mod source

Suor:
Usage:

```squirrel
mod.Mod.Registry.addModSource(
    ::MSU.System.Registry.ModSourceDomain.GitHubTags,
    // A link into a subdir here, but repo itself is also allowed
    "https://github.com/Suor/battle-brothers-mods/tree/master/autopilot"
    // tag prefix, for i.e. tag autopilot-new-2.2.0 be 2.2.0 release
    {Prefix = "autopilot-new-"});
mod.Mod.Registry.setUpdateSource(::MSU.System.Registry.ModSourceDomain.GitHubTags);

// Can still add this for icon with a link
mod.Mod.Registry.addModSource(
    ::MSU.System.Registry.ModSourceDomain.NexusMods,
    "https://www.nexusmods.com/battlebrothers/mods/675");
```

### DIFF
--- a/msu/systems/registry/mod_sources/mod_source_githubtags.nut
+++ b/msu/systems/registry/mod_sources/mod_source_githubtags.nut
@@ -1,0 +1,29 @@
+::MSU.Class.ModSourceGitHubTags <- class extends ::MSU.Class.ModSource
+{
+	static ModSourceDomain = ::MSU.Class.RegistrySystem.ModSourceDomain.GitHubTags;
+	static Regex = regexp(@"https://github.com/([-\w]+)/([-\w]+).*");
+	static BadURLMessage = "A link must point into a Github repository, e.g. 'https://github.com/MSUteam/MSU', or a subtree, e.g. 'https://github.com/Suor/battle-brothers-mods/tree/master/autopilot'.";
+	static Icon = "github";
+	// This is one is set via Mod.Registry.addModSource(..., {Prefix = ...}) -> base.constructor()
+	Prefix = ""
+
+	function getUpdateURL()
+	{
+		local capture = this.Regex.capture(this.__URL);
+		local owner = ::MSU.regexMatch(capture, this.__URL, 1);
+		local repo = ::MSU.regexMatch(capture, this.__URL, 2);
+		return format("https://api.github.com/repos/%s/%s/git/matching-refs/tags/%s?per_page=100",
+			owner, repo, this.Prefix)
+	}
+
+	function extractRelease(_data) {
+		local refPrefix = "refs/tags/" + this.Prefix, maxVersion = "0.0.0";
+		foreach (rec in _data) {
+			local version = rec.ref.slice(refPrefix.len());
+			if (!::MSU.SemVer.isSemVer(version)) continue;
+			if (::MSU.SemVer.compareVersionWithOperator(version, ">", maxVersion))
+				maxVersion = version;
+		}
+		return {Version = maxVersion, Changes = ""};
+	}
+}
--- a/msu/systems/registry/load.nut
+++ b/msu/systems/registry/load.nut
@@ -9,6 +9,7 @@ includeFile("registry_mod_addon");
 
 ::MSU.System.Registry <- ::MSU.Class.RegistrySystem();
 ::MSU.System.Registry.addNewModSource(::MSU.Class.ModSourceGitHub);
+::MSU.System.Registry.addNewModSource(::MSU.Class.ModSourceGitHubTags);
 ::MSU.System.Registry.addNewModSource(::MSU.Class.ModSourceNexusMods);
 ::MSU.getMod <- function( _modID )
 {
--- a/msu/systems/registry/mod_sources/mod_source.nut
+++ b/msu/systems/registry/mod_sources/mod_source.nut
@@ -2,11 +2,21 @@
 {
 	static ModSourceDomain = null;
 	static Regex = null;
+	static BadURLMessage = "A link must be pointing to a specific mod.";
+	static Icon = null;
 	__URL = null;
 
-	constructor( _url )
+	constructor( _url, _opts = null )
 	{
+		if (this.Regex != null && !this.Regex.match(_url))
+		{
+			::logError(this.BadURLMessage);
+			throw ::MSU.Exception.InvalidValue(_url);
+		}
 		this.__URL = _url;
+		if (_opts != null) {
+			foreach (key, value in _opts) this[key] = value;
+		}
 	}
 
 	function getUpdateURL()
@@ -17,5 +27,9 @@
 	function getURL()
 	{
 		return this.__URL
+	}
+
+	function extractRelease(_data) {
+		return null;
 	}
 }
--- a/msu/systems/registry/mod_sources/mod_source_github.nut
+++ b/msu/systems/registry/mod_sources/mod_source_github.nut
@@ -2,20 +2,16 @@
 {
 	static ModSourceDomain = ::MSU.Class.RegistrySystem.ModSourceDomain.GitHub;
 	static Regex = regexp("https:\\/\\/github\\.com\\/([-\\w]+)\\/([-\\w]+)");
-
-	constructor( _url )
-	{
-		if (!this.Regex.match(_url))
-		{
-			::logError("A GitHub link must be a link to a specific repository, e.g. 'https://github.com/MSUTeam/MSU' Check to make sure there's not an issue with your URL and that it is formatted the same way as the MSU URL.");
-			throw ::MSU.Exception.InvalidValue(_url);
-		}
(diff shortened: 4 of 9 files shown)


msu/systems/registry/mod_sources/mod_source_githubtags.nut
```diff
@@ -0,0 +1,29 @@
+::MSU.Class.ModSourceGitHubTags <- class extends ::MSU.Class.ModSource
+{
+	static ModSourceDomain = ::MSU.Class.RegistrySystem.ModSourceDomain.GitHubTags;
+	static Regex = regexp("https:\\/\\/github\\.com\\/([-\\w]+)\\/([-\\w]+)/?.*");
+	static BadURLMessage = "A link must be a link into a Github repository repository, " +
+						   "e.g. 'https://github.com/Suor/battle-brothers-mods/something/else/'";
```
Enduriel, line 6: shouldn't be split up

Enduriel:
Could you also mock up some doc for this, so that we could see how it should be implemented and so that it could be easily copied into the MSU wiki later?

msu/systems/registry/mod_sources/mod_source_githubtags.nut
Suor: Fixed

Suor:
> Could you also mock up some doc for this, so that we could see how it should be implemented and so that it could be easily copied into the MSU wiki later?

Not sure what do you mean. I have an example in commit message above, can copy that to Wiki. 

Or you mean a way to implement an extra type of source? I don't know a way to document it concise enough so that it will be more useful than reading actual source code :)

Suor:
Maybe change this:

```squirrel
	function addModSource( _domain, _url )
	{
		::MSU.requireOneFromTypes(["string", "table"], _url);
		this.__ModSources[_domain] <- ::MSU.System.Registry.ModSources[_domain](_url);
	}
```

to

```squirrel
	function addModSource( _domain, _url, _opts = null )
	{
        ::MSU.requireString(_url);
        if (_opts != null) ::MSU.requireTable(_opts);
		this.__ModSources[_domain] <- ::MSU.System.Registry.ModSources[_domain](_url, _opts);
	}
```

?

Enduriel:
I mean the method of implementing the additional source, this would have to be documented in the MSU wiki as that is standard practice for MSU, and it's supposed to make it more accessible. We don't really point people at the source code, an explanation + an implementation example is required :/.

Suor:
> I mean the method of implementing the additional source, this would have to be documented in the MSU wiki as that is standard practice for MSU, and it's supposed to make it more accessible.

Heh, do you mean implementing inside or outside MSU? And someone adding this simply from docs is just not realistic :). Say you have docs for custom mod settings items, but there is no way I could have done that not reading source throughly. I will try to come up with something though.

Suor:
Ok, added the suggested change.

Enduriel:
I think custom mod settings is a fair example, but admittedly that stuff is pretty complex and requires working with a lot of existing systems within MSU. The GitHub tags shouldn't actually be that complicated to implement from what I understand, and doesn't require interfacing with any internal stuff, so I don't think it should be that problematic?

Suor:
Ok, here it is:

To add new mod source one needs to do several things:

1. Add a new item to `::MSU.System.Registry.ModSourceDomain` enum.
2. Implement `::MSU.Class.ModSource` descendant.
3. Register it with a `::MSU.System.Registry.addNewModSource()` call.
4. *(optionally)* Add `Regex` and `BadURLMessage` to restrict allowed urls.
5. *(optionally)* Add `Icon` to show icon link to a mod on your source. 
6. *(optionally)* Provide `.getUpdateURL()` and `.extractRelease()` methods to support update checks.

Here is a minimal example needed to add a new source, I am providing a way to add links to 
a non-existent hacky-stash.game website here:

```squirrel
::MSU.System.Registry.ModSourceDomain.add("HackyStash");
::YourNamespace.ModSourceHackyStash <- class extends ::MSU.Class.ModSource
{
    static ModSourceDomain = ::MSU.Class.RegistrySystem.ModSourceDomain.HackyStash;
}
::MSU.System.Registry.addNewModSource(::HackMSU.ModSourceHackyStash);
```

Let's add URL validation and some icon:

```squirrel
::YourNamespace.ModSourceHackyStash <- class extends ::MSU.Class.ModSource
{
    ...
    static Regex = regexp(@"https://hacky-stash.game/mods/([-\w]+)");
    static BadURLMessage = "A link must point into directly to a mod page, e.g. 'https://hacky-stash.game/mods/cleaver-axe-overhaul'";
    static Icon = "github";  // Only github or nexusmods images go with MSU
}
```

Now any typos and other mishaps in the URL will be caught, like people typing in `https://stashy-hack.mage/mod/more-zombies` or something. Also an icon link to hacky-stash.game will show up as long as a mod author will provide another way of checking for updates.

To enable update checking for our mod source we will need to add two extra methods:

```squirrel
::YourNamespace.ModSourceHackyStash <- class extends ::MSU.Class.ModSource
{
    ...
    function getUpdateURL()
    {
        local capture = this.Regex.capture(this.__URL);
        local slug = ::MSU.regexMatch(capture, this.__URL, 1);
        // Forturnately hacky slash provides an API
        return format("https://api.hacky-stash.game/mods/%s/releases", slug)
    }

    function extractRelease(_data) {
        // we get a JSON with with all releases, but the latest one goes first,
        if (_data.len() == 0) return null; // protect against no releases
        local release = _data[0];
        // For MSU to understand us we need to send a table with 2 keys:
        //     Version - a semver string of latest available release
        //     Changes - a description of changes brought by it, pass empty string if absent 
        return {Version = release.version, Changes = release.message || ""}
    }
}
```

Enduriel:
Thank you, this is all pretty great, one final thing that I'd ask that will be pretty annoying, but could you separate the `refactor` (cleaning up the way mod sources are defined) and the `feat` (adding the github tags mod source) into 2 separate commits? Once that's done I'd be happy to approve, really like this.

Suor:
Ok, done.

Enduriel (approved):

msu/systems/registry/mod_sources/mod_source_githubtags.nut
```diff
@@ -0,0 +1,28 @@
+::MSU.Class.ModSourceGitHubTags <- class extends ::MSU.Class.ModSource
+{
+	static ModSourceDomain = ::MSU.Class.RegistrySystem.ModSourceDomain.GitHubTags;
+	static Regex = regexp("https:\\/\\/github\\.com\\/([-\\w]+)\\/([-\\w]+)/?.*");
+	static BadURLMessage = "A link must point into a Github repository repository, e.g. 'https://github.com/Suor/battle-brothers-mods/something/else/'";
+	static Icon = "github";
+	Prefix = ""
+
```
TaroEld, line 8: The constructor based on the previous _opts comments would be something like this:
```
	constructor( _url, _opts = null )
	{
		base.constructor(_url, _opts);
		if (_opts != null && "Prefix" in _opts)
		{
			this.Prefix = _opts.Prefix;
		}
	}
```

msu/systems/registry/mod_sources/mod_source_githubtags.nut
```diff
@@ -0,0 +1,28 @@
+::MSU.Class.ModSourceGitHubTags <- class extends ::MSU.Class.ModSource
+{
+	static ModSourceDomain = ::MSU.Class.RegistrySystem.ModSourceDomain.GitHubTags;
+	static Regex = regexp("https:\\/\\/github\\.com\\/([-\\w]+)\\/([-\\w]+)/?.*");
```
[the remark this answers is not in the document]
Suor: I checked and it does match, because there is no need to escape /, but may change to @ and never escape it for consistency

msu/systems/registry/mod_sources/mod_source_githubtags.nut
Suor: Then one will pass `{prefix = "autopilot-new-"}` and will happily debug it for an hour instead of getting an error instantly now. Also will need to repeat for each key. 

Why fight against the language instead of using it?

TaroEld: You can still have erroring if you want. Yes, you will repeat for each key, which amounts to one key at this time, and makes it clear that this key is set in this space for this one mod source.

TaroEld: Okay look, it's fine for me, just add some clarifying comment on the Prefix member that it's set via the _opts parameter in the base class.

Suor: Added a comment.

TaroEld (approved):

PR #430, opened 2024-09-28
feat: add unique id to entity that is serialized

LordMidas:
Vanilla has a `getID()` for entities but that ID is changed on every game load. This PR adds a Unique ID to every entity that is preserved across game save/load. This can be useful for a number of things.

### DIFF
--- a/msu/hooks/entity/tactical/entity.nut
+++ b/msu/hooks/entity/tactical/entity.nut
@@ -1,34 +1,21 @@
 ::MSU.MH.hook("scripts/entity/tactical/entity", function(q) {
-	q.m.UID <- 0;
-
 	q.create <- @(__original) function()
 	{
 		__original();
-		this.m.UID = ::MSU.Utils.generateUID();
+		this.getFlags().set("MSU_UID", ::MSU.Utils.generateUID());
 	}
 
 	q.getUID <- function()
 	{
-		return this.m.UID;
-	}
-
-	q.onSerialize = @(__original) function( _out )
-	{
-		this.m.Flags.set("MSU_UID", this.m.UID);
-		__original(_out);
+		return this.getFlags().get("MSU_UID");
 	}
 
 	q.onDeserialize = @(__original) function( _in )
 	{
 		__original(_in);
-		if (this.getFlags().has("MSU_UID"))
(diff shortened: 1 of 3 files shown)


TaroEld:
Don't entities already have IDs?
actor.getID() certainly is used a lot.

LordMidas:
> Don't entities already have IDs? actor.getID() certainly is used a lot.

They are not persistent and are changed upon every game load.

Darxo (approved):

msu/hooks/entity/tactical/entity.nut
```diff
@@ -0,0 +1,24 @@
+::MSU.MH.hook("scripts/entity/tactical/entity", function(q) {
+	q.create = @(__original) function()
+	{
+		__original();
+		this.getFlags().set("MSU_UID", ::MSU.Serialization.isLoading() ? null : ::MSU.Utils.__generateUID());
```
TaroEld: This should probably still be in a deliberate `setUID()` function to make sure (as much as possible) that we always generate a new UID if we set one.

msu/utils/utils.nut
```diff
@@ -244,6 +245,12 @@
 		}
 	}
 
+	// Private
+	function __generateUID()
+	{
+		return ::MSU.Utils.UID++;
+	}
+
```
TaroEld, line 253: Should this be private? Should other mods not also be able to use this?

LordMidas: This is just for safety. I want to make sure that it isn't used in any way that can lead to the whole system being messed up. For example, someone uses it during ser/de in a way that breaks the UIDs. Or maybe something else idk.

msu/hooks/entity/tactical/entity.nut
LordMidas: I don't think that's important because this `create` and `onDeserialize` below are the only two places where it's gonna be set. Nevertheless, I've pushed a commit implementing your suggestion - take a look at that. But I think we don't need to do this and should probably leave it as I did it first.

msu/utils/utils.nut
TaroEld: ok, keep it like that for now I suppose.

TaroEld (approved):

PR #434, opened 2024-10-06
feat: add new onAnyBroHired event for skills

Darxo:
`onHired` is an event already present in Vanilla but only in the `player.nut` and for origins.
Skills however can't react to the hiring of someone.

~~### This PR proposes two new events for skills:~~
```
// This event is triggered once for every skill of a player character directly after the onHired event of that player class is triggered
function onHired()

// This event is triggered once for every skill of every other character in the player roster after the onHired event of a player class is triggered~~
function onOtherPlayerHired( _otherCharacter )
```
### This PR proposes one new events for skills:
```
/// This event is triggered once for every skill of every character in the player roster after the onHired event of a player class is triggered
/// @param _bro is a refernce to the player object whose onHired event was triggered
function onAnyBroHired <- function( _bro )
```

## Possible Use Cases
~~### onHired~~
### onAnyBroHired
- Certain backgrounds can apply mood buff/debuff on hiring, depending on what other backgrounds are already present in the player party
- Some traits might act as a wildcard/unrevelaed at first (despite trying out) and only transform into a real trait after hiring

~~### onOtherPlayerHired~~
- Certain backgrounds might cause a brother to get a positive/negative mood when you hire a certain background
- Certain backgrounds/perks could grant a bonus to every new recruit (e.g. free perk/stats)

### DIFF
--- a/msu/hooks/entity/tactical/player.nut
+++ b/msu/hooks/entity/tactical/player.nut
@@ -27,6 +27,16 @@
 		this.m.LevelUpsSpent++;
 	}
 
+	q.onHired = @(__original) function()
+	{
+		__original();
+
+		foreach (bro in ::World.getPlayerRoster().getAll())
+		{
+			bro.getSkills().onAnyBroHired(this);
+		}
+	}
+
 	q.onSerialize = @(__original) function( _out )
 	{
 		this.getFlags().set("LevelUpsSpent", this.m.LevelUpsSpent);
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -437,6 +437,12 @@
 
 		return _tooltip;
 	}
+
+	/// This event is triggered once for every skill of every character in the player roster after the onHired event of a player class is triggered
+	/// @param _bro is a reference to the player object whose onHired event was triggered
+	q.onAnyBroHired <- function( _bro )
+	{
+	}
 });
 
 ::MSU.QueueBucket.VeryLate.push(function() {
--- a/msu/hooks/skills/skill_container.nut
+++ b/msu/hooks/skills/skill_container.nut
@@ -538,4 +538,13 @@
 
 		return ret;
 	}
+
+	/// This event is triggered once for every other character in the player roster after the onHired event of a player class is triggered
+	/// @param _bro is a reference to the player object whose onHired event was triggered
+	q.onAnyBroHired <- function( _bro )
+	{
+		this.callSkillsFunction("onAnyBroHired", [
+			_bro
+		]);
+	}
 });


TaroEld (approved):
LGTM

Enduriel (approved):

msu/hooks/entity/tactical/player.nut
```diff
@@ -27,6 +27,21 @@
...
+
+		foreach (bro in ::World.getPlayerRoster().getAll())
+		{
+			if (bro.getID() != this.getID())
+			{
+				bro.getSkills().onOtherPlayerHired(this);
+			}
+		}
+
+	}
```
LordMidas: Note that starting characters in an origin do not call this `player.onHired()` function. So, should these events be called for those characters?

msu/hooks/skills/skill.nut
```diff
@@ -437,6 +437,17 @@
...
+// New Events
+	// This event is triggered once for every skill of a player character directly after the onHired event of that player class is triggered
+	q.onHired <- function()
+	{
+	}
+
+	// This event is triggered once for every skill of every other character in the player roster after the onHired event of a player class is triggered
+	q.onOtherPlayerHired <- function( _otherCharacter )
+	{
+	}
```
LordMidas: Two points:
- Would it be a better setup is to simply have a single event `onHired( _bro )` or `onAnyBroHired( _bro )`? Then for the case of the bro himself you pass him, and for others you pass the other actor.
- The word `Player` in the function name `onOtherPlayerHired` seems awkward as it has other connotations besides the `player.nut` class. So maybe something like `onOtherBroHired` is better. Other suggestions?

msu/hooks/skills/skill_container.nut
```diff
@@ -538,4 +538,19 @@
 
 		return ret;
 	}
+
+// New Events
```
LordMidas: This `// New Events` comment is completely out of place here as there are many new events even above this. So I'd just remove this single line.

msu/hooks/skills/skill.nut
```diff
@@ -437,6 +437,17 @@
 
 		return _tooltip;
 	}
+
+// New Events
```
LordMidas: Same comment about this `// New Events` line.

LordMidas (changes requested):

msu/hooks/entity/tactical/player.nut
Darxo: For non-random scenarios it is up to the scenario designer to manually implement interactions that his starting brothers might have.
For this implementation im purely leeching onto the vanilla onHired function and keeping within its juristiction

msu/hooks/skills/skill.nut
Darxo: I switched to the single event

msu/hooks/entity/tactical/player.nut
```diff
@@ -27,6 +27,16 @@
 		this.m.LevelUpsSpent++;
 	}
 
+	q.onHired = @(__original) function()
+	{
+		__original();
+
+		foreach (bro in ::World.getPlayerRoster().getAll())
+		{
+			bro.getSkills().onAnyoneHired(this);
```
TaroEld: onAnyoneHired here, onAnyBroHired later.
Personally I also think that onBroHired is good enough.

msu/hooks/skills/skill.nut
```diff
@@ -437,6 +437,12 @@
 
 		return _tooltip;
 	}
+
+	/// This event is triggered once for every skill of every character in the player roster after the onHired event of a player class is triggered
+	/// @param _bro is a refernce to the player object whose onHired event was triggered
```
TaroEld: refernce -> reference

TaroEld (approved):
LGTM after typo fixes / name change.

msu/hooks/entity/tactical/player.nut
Darxo: I removed the left-over `onAnyBroHired` here.
If you prefer another term over `onAnyBroHired` you gotta convince midas

LordMidas (approved):

msu/hooks/entity/tactical/player.nut
LordMidas: Ok, I think we can do it like this. If there are concerns in the future we can resolve them at that time.

PR #438, opened 2025-02-05
feat: add functions to check if game is being saved or loaded

### LINKED ISSUE #437: [REQUEST] add global IsDuringSerialize and IsDuringDeserialize flags
Some functionality or events should probably not trigger during deserialization.
I don't have an example just yet. Initially I thought these flags might solve one of my problems but that was not the case.

Hook the `world_state.nut` -> `onDeserialize` and `onSerialize` functions very late and toggle two new flags for them respectively:
- `::MSU.System.Serialization.IsDuringOnSerialize`
- `::MSU.System.Serialization.IsDuringOnDeserialize`

They should behave similar to the already existing `::MSU.System.Serialization.IsDuringOnBeforeSerialize` flag

LordMidas:


msu/utils/serialization.nut
```diff
@@ -16,6 +16,18 @@
...
+
+	function isSaving()
+	{
+		return this.IsDuringSerialization;
+	}
+
+	function isLoading()
+	{
+		return this.IsDuringDeserialization;
+	}
```
LordMidas: Should this be added here or in `world_state`?

Darxo: in `serialization.nut` the name `isLoading` is less ambiguous.

One might also consider the complete loading-screen between combat and world map to be "loading", even though there is no serialization happening there, other than a short autosave

LordMidas: I was hoping to have shorter names than `IsDuringSerialization`. Is this setup clear enough with `::MSU.Serialization.isSaving()` and `::MSU.Serialization.isLoading()`? Or is it important to use the `serializing` and `deserializing` nomenclature?

TaroEld: With it being in the Serialization namespace, I think that should be rather clear.

TaroEld (approved):

Enduriel (approved):

PR #439, opened 2025-02-05
fix: only fire skill execution events if skill actually executed

### LINKED ISSUE #422: [BUG] onBeforeAnySkillExecuted and onAfterAnySkillExecuted do not guarantee that the skill was actually executed
### Versions
 - MSU: 1.6.0

### Describe the bug
`onBeforeAnySkillExecuted` is a new event introduced by MSU
It's described as
> The first function is called before the used skill's use function triggers, whereas the second function is called after the use function is complete.

That is correct with its implementation.

From the name itself one would assume that this only triggers when the `skill` in question is guaranteed to be used afterwards. It is "about to be executed" afterall.
In Reality there are still some checks in the vanilla `use` function which can prevent the skill from actually being executed.
Like if the skill is not actually affordable or the targeting does not check out.

Now the problem is that mods using this event can't know if the skill in question is actually used for sure. 
I would even go so far as to say there is no Event in Vanilla/MSU that guarantees fire when a skill is executed (it's `onUsed` is called). Currently we can only rely on the two MSU functions and those have false positives

### Expected behavior
The `onBeforeAnySkillExecuted` event should only fire when the skill is really executed and not cancelled last second (red arrow)
The `onAfterAnySkillExecuted` event should only fire when the skill is really executed and not cancelled last second (green arrow)

image

### Possible Solution

Do a hookTree with the latest possible timing around the `onUse` function of `skill.nut`.
Then you guarantee that the skill actually executed.

However some mods might hard-call this function. Then MSU would fire new types of false-positves. So this solution still needs some code around the initial `use` function flipping some variable.

### Alternative I considered

Do a normal hook of the `setPreviewSkillID` function from `actor.nut` and check the stack whether it was called by `use`. If so then we trigger `onBeforeAnySkillExecuted` and set a global `skillActuallyExecuted` variable to `true`.
Then after `use` from `skill.nut` ended, we check whether our global variable `skillActuallyExecuted` is true. If it is, we trigger `onAfterAnySkillExecuted`

LordMidas:
Vanilla `use` function has checks built in which may cause it to return early and not actually execute the skill. Hooking onUse ensures that we only fire it when the skill is being executed, however in this case we still need a flag to switch from `use` as `onUse` can be called manually as well.

This will require us to update our documentation as it currently says that these events are fired before and after `use`. So now we'd have to say they fire before and after `onUse`.

### DIFF
--- a/msu/hooks/skills/skill.nut
+++ b/msu/hooks/skills/skill.nut
@@ -17,6 +17,8 @@
 });
 
 ::MSU.MH.hook("scripts/skills/skill", function(q) {
+	q.m.MSU_IsUsingSkill <- false;
+	q.m.MSU_IsUsingSkillForFree <- false;
 	q.m.AIBehaviorID <- null;
 	q.m.DamageType <- null;
 	q.m.ItemActionOrder <- ::Const.ItemActionOrder.Any;
@@ -306,18 +308,10 @@
 
 	q.use = @(__original) function( _targetTile, _forFree = false )
 	{
-		// Save the container as a local variable because some skills delete
-		// themselves during use (e.g. Reload Bolt) causing this.m.Container
-		// to point to null.
-		local container = this.m.Container;
-		local targetEntity = _targetTile.IsOccupiedByActor ? _targetTile.getEntity() : null;
-
-		container.onBeforeAnySkillExecuted(this, _targetTile, targetEntity, _forFree);
-
+		this.m.MSU_IsUsingSkill = true;
+		this.m.MSU_IsUsingSkillForFree = _forFree;
 		local ret = __original(_targetTile, _forFree);
-
-		container.onAnySkillExecuted(this, _targetTile, targetEntity, _forFree);
-
+		this.m.MSU_IsUsingSkill = false;
 		return ret;
 	}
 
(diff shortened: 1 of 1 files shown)


Darxo:
An event for the initial `onBeforeAnySkillExecuted` timing is very much worth keeping.
At this point no Action Points nor Fatigue have yet been spent so this is the perfect event to snapshot those values, if we want to create effects that are based on the amount of AP/Fatigue you have spent, or which reimburse you your spent resources.

With this mods proposal we also have a small issue that there might already be mods out there using `onBeforeAnySkillExecuted` for exactly the above purpose, which would break when MSU moves the timing.
Yet `onBeforeAnySkillExecuted` fits as a name for both timings

LordMidas:
> An event for the initial `onBeforeAnySkillExecuted` timing is very much worth keeping. At this point no Action Points nor Fatigue have yet been spent so this is the perfect event to snapshot those values, if we want to create effects that are based on the amount of AP/Fatigue you have spent, or which reimburse you your spent resources.
> 
> With this mods proposal we also have a small issue that there might already be mods out there using `onBeforeAnySkillExecuted` for exactly the above purpose, which would break when MSU moves the timing. Yet `onBeforeAnySkillExecuted` fits as a name for both timings

Actually your point is valid and this means that this PR would break the current behavior, so we cannot do it like this. Currently `onBeforeAnySkillExecuted` runs before AP/Fatigue is spent, so we'd have to run it earlier. Perhaps one has to go in a very weird way and hook `spawnOverlay` and check stack trace to confirm that it is coming from `use` and if yes then trigger the event. Because that function is called from `use` just before AP/Fatigue is spent. But that seems very ugly.

PR #441, opened 2025-03-04
Feat getKeybind and capitalized key combinations

LordMidas:
- add `getKeybind` function to keybind mods addon
- add `_capitalizeFirst` as optional parameter to `getKeyCombinations` of `abstract_keybind` which, if true, returns first letter capitalized keybind string.

### DIFF
--- a/msu/systems/keybinds/abstract_keybind.nut
+++ b/msu/systems/keybinds/abstract_keybind.nut
@@ -39,6 +39,12 @@
 		return this.KeyCombinations.reduce(@(_a, _b) _a + "/" + _b);
 	}
 
+	function getKeyCombinationsCapitalized()
+	{
+		if (this.KeyCombinations.len() == 0) return "";
+		return this.KeyCombinations.reduce(@(_a, _b) ::MSU.String.capitalizeFirst(_a) + "/" + ::MSU.String.capitalizeFirst(_b));
+	}
+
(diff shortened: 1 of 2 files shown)


msu/systems/keybinds/abstract_keybind.nut
```diff
@@ -33,9 +33,15 @@
...
 	{
 		if (this.KeyCombinations.len() == 0) return "";
+
+		if (_capitalizeFirst)
+		{
+			return this.KeyCombinations.reduce(@(_a, _b) ::MSU.String.capitalizeFirst(_a) + "/" + ::MSU.String.capitalizeFirst(_b));
+		}
+
 		return this.KeyCombinations.reduce(@(_a, _b) _a + "/" + _b);
 	}
```
TaroEld: I'm not a fan of this being a parameter, I'd even prefer it just being another function that does the same thing but with your changes, `getCapitalizedKeyCombinations` or something. At least then we can easily refactor things if necessary.

TaroEld (changes requested):

msu/systems/keybinds/abstract_keybind.nut
LordMidas: Ok I will change it to a separate function.

TaroEld (approved):
at some point we should probably add some validations for our getters

PR #444, opened 2025-03-10
feat: convert floats to ints if they are the same value and should be ints

TaroEld:
See https://github.com/MSUTeam/MSU/pull/412


LordMidas (approved):
This solves the known issue of vanilla writing floats as integers in named_armor with armor attachments.

PR #445, opened 2025-03-12
fix: tooltips having a jagged line in their background-image if the height increases too much

TaroEld:


LordMidas (approved):
I have no idea about this, so you decide. All I can say is that if it is a vanilla issue then please write `VanillaFix` instead of `Fix` in the comment. Similarly in the commit message phrase it as `fix(vanilla):`.

PR #448, opened 2025-03-21
feat: deprecate msu onMovementFinished and adjust for vanilla 1.5.1.4

LordMidas:


msu/hooks/entity/tactical/actor.nut
```diff
@@ -10,9 +10,13 @@
...
-		this.m.IsMoving = true;
-		this.m.Skills.onMovementFinished(_tile);
-		this.m.IsMoving = false;
+		// legacy support for MSU-added skill_container.onMovementFinished for vanilla up to 1.5.0.15
+		if (::Const.Serialization.Version < 65)
+		{
+			this.m.IsMoving = true;
+			this.m.Skills.onMovementFinished();
+			this.m.IsMoving = false;
+		}
```
Enduriel: Actually, I don't think we can do this. I'm pretty sure legends modifies the Serialization version, so we have to use something like `::Hooks.getMod("vanilla").getVersion() <= ::Hooks.SQClass.ModVersion("1.5.0-15")`

msu/hooks/skills/skill_container.nut
```diff
@@ -87,11 +87,35 @@
 		]);
 	}
 
-	q.onMovementFinished <- function( _tile )
+	// Overwrite vanilla function to pass tile for legacy support of MSU-added skill_container.onMovementFinished.
+	// Vanilla added their own onMovementFinished event in 1.5.1.4 without a `_tile` parameter.
+	// Naturally, the MSU-added function with _tile parameter is deprecated and should no longer be used by mods.
+	q.onMovementFinished = @(__original) function()
```
Enduriel: remove `__original`

Enduriel (changes requested):

msu/hooks/entity/tactical/actor.nut
LordMidas: >  legends modifies the Serialization version

That seems like a bad thing to do. But ok.

> ::Hooks.getMod("vanilla").getVersion() <= ::Hooks.SQClass.ModVersion("1.5.0-15")

That seems like quite an overhead on every single `onMovementFinished` call. Maybe we check this during startup and store it in a global boolean somewhere?

Enduriel: I think the overhead is inconsequential since this is on the actor-wide `onMovementFinished` call, and I think this is the cleanest and most readable way to do this.

TaroEld (approved):

PR #456, opened 2025-04-15
feat: add function for stable sorting arrays

LordMidas:
This function seems important because we had a really mysterious bug in Reforged caused by a very sneaky bug in Dynamic Spawns Framework which wasn't really a bug but a consequence of the fact that the squirrel `array.sort` does not preserve order for equal elements. It was causing units of equal cost in a unit block to become randomly ordered and this was not very obvious at all.

We need a function that sorts in a stable way, at least for use cases where the order is important.

### DIFF
--- a/msu/utils/array.nut
+++ b/msu/utils/array.nut
@@ -79,6 +79,32 @@
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


TaroEld (approved):
Pretty niche but sure.
I ran a few test-cases through it, looks fine.

PR #459, opened 2025-04-30
feat: remove overwrite of vanilla skill container events

LordMidas:
- This allows other mods that want to queue before MSU and modify these events to actually work properly with their modifications.
- The main goal of the overwrites, as far as I (LordMidas) understand, was to have a DRY method of applying the `wasUpdating` fix. However, upon reflection I have realized that that fix is only necessary for the `buildPropertiesForXYZ` functions. Because it is these functions which have a chance of being called from within other skill container events. The other events e.g. `onTurnEnd` etc. are very well secluded and are only ever called independently, so never run into an issue of setting `IsUpdating` to false too early. So we keep the `wasUpdating` check manually for the `buildProperties` type functions, but for the others we remove the overwrites.

Eventually with the combination of the other fix in #447 this might as well fix some hidden/unknown bugs being caused by MSU.


TaroEld (approved):
I don't know enough about the skills thing but the code looks OK

PR #461, opened 2025-05-01
feat: add ::MSU.Math.clamp function

Darxo:
This common math function is missing in the squirrel ::Math table.
It is very useful to improve readability, as currently we need to implement clamping by writing something like this:
`return ::Math.max(0, ::Math.min(100, _foo));`

That could now be written by 
`return ::MSU.Math.clamp(_foo, 0, 100);`
And the intent would be clear right away; not half-way into reading the line (well, for bigger lines)

### DIFF
--- a/msu/utils/math.nut
+++ b/msu/utils/math.nut
@@ -1,4 +1,16 @@
 ::MSU.Math <- {
+	// Clamps an integer or float _value to be within [_min, _max] range
+	// throw InvalidValue exception when _min is greater than _max
+	function clamp(_value, _min, _max)
+	{
+		if (_min > _max)
+		{
+			::logError("_min must be less than or equal to _max");
+			throw ::MSU.Exception.InvalidValue(_min);
+		}
+		return _value < _min ? _min : (_value > _max ? _max : _value);
+	}
+
 	// returns log2 of _num, floored (as an int)
 	function log2int( _num )
 	{


msu/utils/math.nut
```diff
@@ -1,4 +1,16 @@
...
+	// throw InvalidValue exception when _min is greater than _max
+	function clamp(_value, _min, _max)
+	{
+		if (_min > _max)
+		{
+			::logError("_min must be less than or equal to _max");
+			throw ::MSU.Exception.InvalidValue(_min);
+		}
+		return _value < _min ? _min : (_value > _max ? _max : _value);
+	}
```
LordMidas, line 12: This is definitely a good feature to add. But this implementation does not preserve the type of `_value`. E.g. `clamp(someFloat, 1, 5)` will now convert the float to an integer because I passed integers as min and max. So:
- Either make it preserve the type i.e. the return value should match the type of `_value`.
- Or make this function **always** return an integer (similar to `::Math.max`) and add a separate function `clampf` for floats which always returns a float (similar to `::Math.maxf`).

Perhaps the first option is better? @TaroEld What do you think?

LordMidas (changes requested):

msu/utils/math.nut
```diff
@@ -1,4 +1,16 @@
 ::MSU.Math <- {
+	// Clamps an integer or float _value to be within [_min, _max] range
+	// throw InvalidValue exception when _min is greater than _max
+	function clamp(_value, _min, _max)
```
LordMidas, line 4: Style wise add a space before after the parameters.

Darxo:
I prefer if one of you just implements this function as you see fit

PR #468, opened 2025-05-12
feat: add clamp function to msu math library

### LINKED ISSUE #464: [REQUEST] Clamp function
The common math function "Clamp" is missing in the squirrel ::Math table.

Currently we need to implement clamping by writing something like this:
`return ::Math.max(0, ::Math.min(100, _foo));`

Right now, especially after you've worked with such lines a bunch of times, you will easily identify it.
But as the line grows larger, because the arguments come from different places, it becomes harder to parse at a glance.

Add a new `::MSU.Math.Clamp` function.

It could look something like this:
```
// Clamps an integer or float _value to be within [_min, _max] range
// throw InvalidValue exception when _min is greater than _max
::MSU.Math.clamp <- function( _value, _min, _max )
{
	if (_min > _max)
	{
		::logError("_min must be less than or equal to _max");
		throw ::MSU.Exception.InvalidValue(_min);
	}
	return _value < _min ? _min : (_value > _max ? _max : _value);
}
```

### Additional context
Here is a now closed PR with some thoughts from the MSU guys about this: https://github.com/MSUTeam/MSU/pull/461

LordMidas:


msu/utils/math.nut
```diff
@@ -79,4 +79,11 @@
 		local tens = ::pow(10, _decimalPlaces);
 		return ::Math.floor(tens * _num) / tens;
 	}
+
+	// The return has the same type as that of _value
+	function clamp( _value, _min, _max )
+	{
+		local ret = _value < _min ? _min : (_value > _max ? _max);
```
TaroEld: Missing : _value after the second ternary.
Needs check if _min > _max, otherwise it will return unexpected values.

TaroEld (changes requested):

TaroEld (changes requested):
Still needs a check for _min > _max. Probably just switch them in that case:
image

TaroEld (approved):

PR #475, opened 2025-11-13
fixed reversed parameter order in onBeforeTargetHit hook

chopeks:


LordMidas (approved):
Thank you for catching and fixing this bug. This is one of those things that doesn't produce an error and just silently never works.

PR #480, opened 2025-12-10
fix: trying to reset fields on potentially unadded skills from weapon

### LINKED ISSUE #479: [BUG] Duplicate Item-Granted Skill skips setContainer, triggering MSU reset warnings
### Versions
 - Game: 1.5.1.7
 - MSU: 1.7.2

### Describe the bug
When an active skill is added by a weapon, but already present on the character, then the `skill::add` function will end very early, resulting in no `setContainer` call for this skill.

This scenario will cause MSU to print a logWarning with a stackstrace, that looks like this:
> MSU Mod resetField("FatigueCost") skill "actives.reload_bolt" does not have base values saved.

That is because MSU only saves the base values during `setContainer`, but it always does a `reset("FatigueCost")` at the end of `addSkill`, no matter if the skill in question was successfullly added to the actor.

### To Reproduce
Steps to reproduce the behavior: (Assumption, as I dont have a game setting without stack-based-skills, which fixes this)
1. Dont use `stack-based_skills`
1. Create or change any weapon so that during `onEquip` it adds the exact same skill twice
2. Add that weapon to your inventory
3. Equip that weapon
4. See the log warning

### Expected behavior
MSU could instead first check, if the `_skill` actually has a container, before calling `reset("FatigueCost")` on it
And then print no warning.

Or MSU prints a warning with better instructions to the modder, what they did wrong

### Additional info

The mod `stack-based-skills` acknowledges this weird behavior and adds some glue code, so it doesn appear:

```js
// We do this so that MSU properly saves this skill's base values
// and doesn't complain when skills added by items are not added to the actor
// thereby not having their base values saved. Because MSU tries to reset certain
// fields in its hook on weapon.addSkill.
_skill.setContainer(this);
_skill.setContainer(null);
```

LordMidas:
Non-stacking skills are not added to a skill_container that already has a skill with that ID. So if a weapon tries to add such a skill, it will not get added to the container. Therefore, we add a guard statement and return early in this case.

### DIFF
--- a/msu/hooks/items/shields/shield.nut
+++ b/msu/hooks/items/shields/shield.nut
@@ -2,6 +2,13 @@
 	q.addSkill = @(__original) function( _skill )
 	{
 		__original(_skill);
+
+		// If the skill wasn't actually added to the container for some reason
+		// (e.g. adding non-stacking skill that already exists in the container)
+		// then we bail out.
+		if (::MSU.isNull(_skill.getContainer()))
+			return;
+
 		if (_skill.isType(::Const.SkillType.Active))
 		{
 			// We reset the FatigueCost so any modifications to it from other skills is reverted
--- a/msu/hooks/items/weapons/weapon.nut
+++ b/msu/hooks/items/weapons/weapon.nut
@@ -22,7 +22,14 @@
 
 	q.addSkill = @(__original) function( _skill )
 	{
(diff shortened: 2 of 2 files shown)


TaroEld (commented):
Should the same not be done for the equivalent shield hook?

TaroEld (approved):

PR #485, opened 2026-02-21
feat: add ability to use event colors to colorize text and values

### LINKED ISSUE #474: [REQUEST] Allow custom positive/negative colors in "colorizeValue" and similar functions

The `::MSU.Text.colorizeValue` function will color the passed value in either green or red, depending on its value.
However, depending on the background color of the textbox in which this value will appear, we might want different shades of green/red that better contrast in those situations.

Vanilla uses `::Const.UI.Color.PositiveValue` and `::Const.UI.Color.NegativeValue` for tooltips.
And it uses `::Const.UI.Color.PositiveEventValue` and `::Const.UI.Color.NegativeEventValue` for Events and Contracts. These variations are brighter and are more readable on the dark brown background.

Support new `kwargs` called something like `UseEventColors = false`.
When someone wants to use the lighter event-colors instead of the tooltip colors, they turn this value to true, when calling any of the MSU coloring functions.

### Describe alternatives you've considered
Support new `kwargs` called something like `CustomPositiveColor = ""` and `CustomNegativeColor = ""`, where the user passes their preferred colors.
The first solution however is much shorter on the user site

LordMidas:


TaroEld (approved):
It's honestly not a really clean solution, but these functions are weird to me anyways so OK I suppose.

PR #487, opened 2026-04-07
feat: add toPrecision based on step to avoid floating point weirdness

TaroEld:


LordMidas (approved):
JS is mostly beyond my skillset but if you say it werks then looks good.

PR #490, opened 2026-04-17
feat: validate mods compatibility when trying to load saved games

### LINKED ISSUE #233: [REQUEST] A system for mods that warns about/prevents loading saves when a added/removed mod would cause issues
Mods can easily break savegames during de-serialization. This happens when they introduce new variables that need saving and loading.
When you remove this mod then that data is still in the save file but gets interpreted as something else.

When I make a mod I should be able to declare it as
- Requires new savegame
and/or
- Corrupts savegames upon removal
And then MSU automatically prevents the player from starting savegames if the condition from above would mean loading them would cause issues.

If a mod only sometimes causes issues then they should not use the feature above. That's only meant for mods that 95%+ cause issues

**Describe alternatives Enduriel considered**
> I think we should have a popup with an explanation of why it's not loading
> And probably allow them to bypass it

LordMidas:
- We save the mod compatibility data for registered mods using the same technology that is available in Modern Hooks.
- We then validate the saved mods just like Modern Hooks validates installed mods.
- We allow mods to adjust the saved mods' compatibility data during validation in case they want to act as a compatibility patch or fork.

## Usage
This system is meant to seamlessly blend with Modern Hooks. Use Modern Hooks and declare your requirements/conflicts as normal. This info will be saved to the save file and will be automatically used to validate compatibility.

### Declaring your mod as unsafe to add to existing saves
Use Modern Hooks and require your own mod i.e.
```squirrel
HooksMod.require("my_own_mod_id");
```

### Declaring your mod incompatible with older versions of itself
Use Modern Hooks and require a version of your mod:
```squirrel
HooksMod.conflictWith("my_own_mod_id < 0.5.5");
```

### Applying a compatibility patch to remove declared incompatibilities
During savegame compatibility validation, you can adjust the CompatibilityData stored in the save file. For this, you need to use the MSU Serialization system. This can be useful when creating compatibility patches or creating forks of mods with different ids and declaring yourself compatible with save games that require the original mod.
```squirrel
// Let's say the save file has `mod_a` which requires `mod_b` version 1.8.0 or greater.
// You made a compatibility patch mod which can drop this requirement, or reduce the requirement to 1.5.0.
MyMSUMod.Serialization.onValidateSavedMods( _modsInfo )
{
  if (_modsInfo.hasMod("mod_a")
  {
    // First clear the compatibility data e.g. requirement/conflict stored for `mod_xyz` about `mod_asdf`.
     _modsInfo.clearCompatibilityData("mod_a", "mod_b");
   // Then you can add new requirement/conflict e.g.
    _modsInfo.getMod("mod_a").require("mod_b >= 1.5.0");
  }
```

image

image

### DIFF
--- a/msu/systems/serialization/saved_mods_info.nut
+++ b/msu/systems/serialization/saved_mods_info.nut
@@ -29,7 +29,7 @@
 			return;
 		}
 
-		foreach (id in split(ids, this.ModIDsSeparator))
+		foreach (id in ::MSU.String.split(ids, this.ModIDsSeparator))
 		{
 			this.Mods[id] <- this.__getModFromInfoString(_metadata.getString(this.MetaDataSavedInfoPrefix + id));
 		}
@@ -41,7 +41,7 @@
 		if (!_metadata.hasData("MSU.SavedModIDs"))
 			return;
 
-		local ids = split(_metadata.getString("MSU.SavedModIDs"), ",");
+		local ids = ::MSU.String.split(_metadata.getString("MSU.SavedModIDs"), ",");
 		foreach (id in ids)
 		{
 			this.Mods[id] <- ::Hooks.SQClass.Mod(id, _metadata.getString(id + "Version") == "" ? "1.0.0" : _metadata.getString(id + "Version"), "", this.EmptyTable);
@@ -56,7 +56,7 @@
 	// Return example: "mod_msu >= 1.8.0"
 	function __getCompatString( _str )
 	{
-		local info = split(_str, this.CompatInfoSeparator);
+		local info = ::MSU.String.split(_str, this.CompatInfoSeparator);
 		local operator = info[2];
 		local version = info[3];
 		local name = info[1];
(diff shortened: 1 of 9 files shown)


msu/systems/serialization/saved_mods_info.nut
```diff
@@ -0,0 +1,230 @@
...
+		{
+			this.__loadOldData(_metadata);
+			return;
+		}
+
+		foreach (mod in split(info, this.ModSeparator))
+		{
+			local info = split(mod, this.ModInfoSeparator);
+			local mod = ::Hooks.SQClass.Mod(info[0], info[2], info[1], this.EmptyTable);
+			if (info[3] != "x")
```
Darxo: The `3` should be a static class variable with a name. Especially since you later use a `[3]` magic number a second time where it is not clear if both of those values mean the same thing

msu/systems/serialization/saved_mods_info.nut
```diff
@@ -0,0 +1,230 @@
...
+			{
+				foreach (req in split(info[3], this.CompatModSeparator))
+				{
+					mod.require(this.__getCompatString(req));
+				}
+			}
+			if (info[4] != "x")
```
Darxo: `4` should be a static class variable

Darxo (changes requested):

Darxo:
Perhaps it's possible to adjust the vanilla warning message that's displayed on those incompatible savegames.
Make it include "Mods" too.
`- Incompatible Version or DLC/Mods Missing -`

image

PR #503, opened 2026-06-28
Add labelInTitle option to SliderSetting

Suor:
Any labels longer then several symbols overlap with a control to the right (if slider is in the left column). This is fine for most cases, where labels is a number or a single short word, but lacks flexibility.

This PR adds an optional `labelInTitle` mode to `SliderSetting`. When enabled, the slider's current-value label is rendered in the setting's title row (right-aligned) instead of next to the slider track.

## API

```squirrel
local setting = ::MSU.Class.SliderSetting(...);
setting.setLabelInTitle(true);
```

I tried passing it to constructor, it works but cumbersome without named params, which squirrel lacks and MSU doesn't emulate. Settled on accessor. 

## Changes

- `slider_setting.nut`: new `LabelInTitle` field (default `false`), `setLabelInTitle(_value)` setter, and `labelInTitle` passed through `getUIData()`.
- `slider_setting.js`: when `labelInTitle` is set, detach the label and append it to the `titleContainer`.

Default behavior is unchanged (`LabelInTitle = false`).

### DIFF
--- a/msu/systems/mod_settings/elements/slider_setting.nut
+++ b/msu/systems/mod_settings/elements/slider_setting.nut
@@ -2,6 +2,7 @@
 {
 	Values = null;
 	Labels = null;
+	LabelInTitle = false;
 	static Type = "Slider";
 
 	constructor( _id, _value, _values, _labels = null, _name = null, _description = null )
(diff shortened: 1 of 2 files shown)


Suor:
Any moves on this?

PR #514, opened 2026-08-09
fix: mod setting lock lost when loading a save

Suor:
A lock a mod puts on a setting at load time now survives a campaign saved before that lock existed.

### DIFF
--- a/msu/systems/mod_settings/abstract_setting.nut
+++ b/msu/systems/mod_settings/abstract_setting.nut
@@ -171,9 +171,12 @@
 
 	function __setFromSerializationTable( _table )
 	{
+		// A lock put on by mod code describes the current mod set, not the save, so outlive it
+		local locked = this.isLocked(), lockReason = this.getLockReason();
 		this.unlock();
 		this.set(_table.Value, true, false, true, true);
-		if (_table.Locked) this.lock(_table.LockReason);
+		if (locked) this.lock(lockReason);
+		else if (_table.Locked) this.lock(_table.LockReason);
 	}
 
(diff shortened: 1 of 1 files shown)


msu/systems/mod_settings/abstract_setting.nut
```diff
@@ -171,9 +171,12 @@
 
 	function __setFromSerializationTable( _table )
 	{
+		// A lock put on by mod code describes the current mod set, not the save, so outlive it
+		local locked = this.isLocked(), lockReason = this.getLockReason();
 		this.unlock();
 		this.set(_table.Value, true, false, true, true);
-		if (_table.Locked) this.lock(_table.LockReason);
+		if (locked) this.lock(lockReason);
+		else if (_table.Locked) this.lock(_table.LockReason);
```
LordMidas, line 179: The pre-existing lock from the save should be prioritized over the current new lock.
```suggestion
		if (_table.Locked) this.lock(_table.LockReason);
		else if (locked) this.lock(lockReason);
```

LordMidas (changes requested):

msu/systems/mod_settings/abstract_setting.nut
Suor: I don't understand that. Why locks are even saved into saves? Also, even if that is ok, the new reason is correct - the mod might have changed, the old is stale

LordMidas: Yeah I think you're right.

LordMidas (approved):

